PR #13, opened 2017-02-04
Fix build of priority benchmark

bergmark:
```
bench-priority/Main.hs:8:1: error:
    Failed to load interface for ‘System.Random’
    It is a member of the hidden package ‘random-1.1’.
    Perhaps you need to add ‘random’ to the build-depends in your .cabal file.
    Use -v to see a list of the files searched for.
```

### DIFF
--- a/http2.cabal
+++ b/http2.cabal
@@ -284,6 +284,7 @@ Benchmark priority
                       , hashtables
                       , heaps
                       , mwc-random
+                      , random
                       , psqueues
                       , stm
 


kazu-yamamoto:
Thank you for this report.

I would like to remove `random` since it causes an error on Travis CI.
I would like to use `mwc-random` instead.
But now I'm waiting for https://github.com/hvr/cassava/issues/129 to install benchmark environment with GHC 8.0.2.

kazu-yamamoto:
This has been fixed in c3f782dd4e843abe1e07f966327bbbda58b362f0.

PR #21, opened 2021-02-10
Support bytestring-0.11

Bodigrim:
Similar to https://github.com/kazu-yamamoto/network-byte-order/pull/2, If this looks too involved, alternatively you can just drop support of GHC 7.10 with a revision: for GHC 8.0+ `bytestring-0.11` provides a backward compatible pattern synonym `PS`. **Upd.:** But in such scenario earlier releases need to be revised with `bytestring < 0.11`, otherwise `cabal` will choose to build them and choke. 

It would be nice to share `withBS` between `Network.HPACK.Huffman.ByteString` and `Network.HPACK.Token`, but I'm not sure where to put such compatibility module.

### DIFF
--- a/Network/HPACK/Huffman/ByteString.hs
+++ b/Network/HPACK/Huffman/ByteString.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE ForeignFunctionInterface #-}
 
 module Network.HPACK.Huffman.ByteString (
@@ -6,6 +7,7 @@ module Network.HPACK.Huffman.ByteString (
   ) where
 
 import Foreign.C.Types (CSize(..))
+import Foreign.ForeignPtr (ForeignPtr)
 import Foreign.Ptr (Ptr, plusPtr)
 import Foreign.Storable (peek)
 import System.IO.Unsafe (unsafeDupablePerformIO)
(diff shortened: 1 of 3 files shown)


kazu-yamamoto:
@Bodigrim Should I do the same thing as `fast-logger`?

Bodigrim:
Yes, please.

kazu-yamamoto:
Done. Would you check?

Bodigrim:
Perfect, thanks!

kazu-yamamoto:
@Bodigrim Are there any issues which I can help?

Bodigrim:
I did not review thoroughly everything; I guess the latest changes should cut off the majority of invalid build plans already.

One random observation: https://github.com/kazu-yamamoto/http-date/blob/24d1a00435a6c31f474503bf3754fa1bffe7b9bb/Network/HTTP/Date/Formatter.hs#L80 need to use http://hackage.haskell.org/package/bytestring-0.11.1.0/docs/Data-ByteString-Internal.html#v:toForeignPtr instead of pattern-matching. This is backwards compatible, no CPP needed.

kazu-yamamoto:
@Bodigrim Thank you for pointing out. I would like to drop the support of GHC 7 in `http-date`. In this case, should I do the same thing as others to `http-date`?

Bodigrim:
Yes, I think so.

Issue #16, opened 2019-02-13
Do not allow update max dynamic table size to 0


swamp-agr:
Hi @kazu-yamamoto,

**Problem statement**

During invocation of `http2-client` that uses `http2` under the hood I encountered `TooSmallTableSize` exception. I create a new issue here to bring some light on this topic. Initial discussion with `http2-client` is there: https://github.com/lucasdicioccio/http2-client/issues/65

**Details**

- server side: `nginx 1.14.0` with `http2` and `https` enabled by default.
- client: `http2-client` with the latest version customized to use `http2` from `master` branch of this repo.
- `:method`: `GET`
- Sequence of frames (trace from `http2-client`):

```
(2019-02-13 14:57:11.820714 UTC,[(SettingsMaxFrameSize,16384),(SettingsMaxConcurrentStreams,100),(SettingsMaxHeaderBlockSize,1048576),(SettingsInitialWindowSize,2147418112),(SettingsEnablePush,1),(SettingsHeaderTableSize,65536)])
(">>> ",0,[(0,SettingsFrame [(SettingsMaxFrameSize,16384),(SettingsMaxConcurrentStreams,100),(SettingsMaxHeaderBlockSize,1048576),(SettingsInitialWindowSize,2147418112),(SettingsEnablePush,1),(SettingsHeaderTableSize,65536)])])
      Table size: 0/655360
("<<< ",0,(FrameHeader {payloadLength = 18, flags = 0, streamId = 0},Right (SettingsFrame [(SettingsMaxConcurrentStreams,128),(SettingsInitialWindowSize,65536),(SettingsMaxFrameSize,16777215)])))
(">>> ",0,[(1,SettingsFrame [])])
      Table size: 0/655360
("<<< ",0,(FrameHeader {payloadLength = 4, flags = 0, streamId = 0},Right (WindowUpdateFrame 2147418112)))
      Table size: 0/655360
[...]

PR #28, opened 2021-06-09
Skip inserting entries that do not fit in the encoding table

### LINKED ISSUE #27: SETTINGS_HEADER_TABLE_SIZE not respected
It seems that http2 servers fail to respect the table size settings sent by the client. This issue manifested for us as a 502 error from an nginx ingress controller in kubernetes, sitting in front of a grpc server written with mu-grpc-server.

To reproduce, I used the example server in the http2 library code:

```haskell
{-# LANGUAGE OverloadedStrings #-}

module Main where

import qualified Control.Exception as E
import Data.ByteString.Builder (byteString)
import Network.HTTP.Types (ok200)
import Network.HTTP2.Server
import Network.Run.TCP (runTCPServer)

main :: IO ()
main = runTCPServer Nothing "50051" runHTTP2Server
  where
    runHTTP2Server s =
      E.bracket
        (allocSimpleConfig s 4096)
        freeSimpleConfig
        (\config -> run config server)
    server _req _aux sendResponse = sendResponse response []
      where
        response = responseBuilder ok200 header body
        header =
          [("Content-Type", "text/plain")]
        body = byteString "Hello, world!\n"
```

and fed it [this](https://gist.github.com/pcapriotti/3f86804e58f70122eed7c1b5c8f523db/raw/649fe15eef5389f4cba8171e838592c5f3c028d2/req) sequence of two requests. The requests contain GRPC invokations, but that is not relevant. There is a SETTINGS_HEADER_TABLE_SIZE frame with value 0 in the beginning, but the response I get for the second request has the following header fragment:

```
20 88 be
```

And `be` decodes to index 62, which is outside the static range (presumably, it is referring to the previously sent content-type header).

[This](https://gist.github.com/pcapriotti/3f86804e58f70122eed7c1b5c8f523db/raw/649fe15eef5389f4cba8171e838592c5f3c028d2/example.pcapng) is a wireshark dump of the interaction.

pcapriotti:
This should fix #27. The new tests show that encoding with a 0-size table prevents the encoder from creating indexed fields that point to entries in the dynamic table.

I also reverted the workaround for #16 introduced in PR #17. Tests still pass.

### DIFF
--- a/Network/HPACK/Table/Dynamic.hs
+++ b/Network/HPACK/Table/Dynamic.hs
@@ -222,7 +222,6 @@ newDynamicTable maxsiz info = do
 
 -- | Renewing 'DynamicTable' with necessary entries copied.
 renewDynamicTable :: Size -> DynamicTable -> IO ()
-renewDynamicTable 0 _ = return () -- FIXME: handle case 'Max table size = 0'.
 renewDynamicTable maxsiz dyntbl@DynamicTable{..} = do
     renew <- shouldRenew dyntbl maxsiz
     when renew $ do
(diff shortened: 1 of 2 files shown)


kazu-yamamoto (approved):
Looks excellent!

kazu-yamamoto:

PR #31, opened 2021-11-25
Return correct status messages in HTTP2 client

pcapriotti:
The `Enum` instance of `Status` can be used to create values containing the usual values for the message. This replaces the hardcoded "fixme" value that was previously returned as part of the status.


kazu-yamamoto (approved):
Excellent!

kazu-yamamoto:
Rebased and merged. Thank you for your contribution!

PR #34, opened 2021-12-13
Avoid empty data

pcapriotti:
This fixes an issue with the client when receiving a streaming response from the server. It can happen that too many empty data frames are sent, causing the client to terminate the connection (as a DoS protection).


kazu-yamamoto (approved):
LGTM.
I confirmed that test fails sometime without the patch for `Sender.hs`.

kazu-yamamoto:
Merged. Thank you for your contribution!

kazu-yamamoto:
I have released a new version.

PR #47, opened 2023-03-03
Implementing streaming flush

### LINKED ISSUE #46: flush of streaming does not put a packet out to a network
The code of `Sender` does not flush a packet when `flush` is called.

kazu-yamamoto:
@epoberezkin  This should fix #46.

This code call `writev(2)` or something for flusing. It is up to the TCP layer how TCP packets are organized.

### DIFF
--- a/Network/HTTP2/Arch/Sender.hs
+++ b/Network/HTTP2/Arch/Sender.hs
@@ -134,9 +134,9 @@ frameSender ctx@Context{outputQ,controlQ,connectionWindow,encodeDynamicTable}
         let payloadOff = off0 + frameHeaderLength
             datBuf     = confWriteBuffer `plusPtr` payloadOff
             datBufSiz  = confBufferSize - payloadOff
-        Next datPayloadLen mnext <- curr datBuf datBufSiz lim -- checkme
+        Next datPayloadLen reqflush mnext <- curr datBuf datBufSiz lim -- checkme
         NextTrailersMaker tlrmkr' <- runTrailersMaker tlrmkr datBuf datPayloadLen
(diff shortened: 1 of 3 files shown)


kazu-yamamoto:
To fix another bug, I have merged this PR.

epoberezkin:
Am I right that #46/7 should increase streaming bandwidth? Will go test - exciting - thank you! :)

epoberezkin:
no, it didn't have any impact on performance - getting 1gb file via xftp CLI/server on localhost is taking about 28 seconds to upload, and 24 seconds to download (including ~2sec to encrypt/decrypt). I think we should be able to get it down to 5-10sec once the streaming is faster.

Stalling that happens in the client with remote servers is still there, but I only updated client code so far - I will re-test when the remote server is updated. It didn't and doesn't happen with the local server.

I think streaming bandwidth the most important thing to improve in the library - to make streaming send frames without waiting for response to each, up to some number of frames (possibly, I am completely wrong about how it works, but I observe that up/download time is hugely impacted by the connection latency, even with very high bandwidth).

But nothing seems to be broken :)

kazu-yamamoto:
After #49 is resolved, I will tackle performance issues.

This PR is already closed.
So, would you create a new issue and copy-and-paste the comment above?

epoberezkin:
> So, would you create a new issue and copy-and-paste the comment above?

Sure!

epoberezkin:
will create two issues!

kazu-yamamoto:
OK. Thanks!

epoberezkin:
Done: #50 and #51

PR #64, opened 2023-04-01
using Word as the internal representation of window size

### LINKED ISSUE #63: Support for 32bit platforms
Client code compiled for ARMv7a (32bit) fails on receiving streamed request bodies (sending mostly works, it only inconsistently gets stuck on 32bit, so sending may also be affected in some cases).

This is quite unfortunate as we just released v4.6 of SimpleX Chat that supports ARMv7a on Android, and v4.6.1 about to be released includes the experimental support for file transfers that uses http2, and v5.0 planned in April was going to switch to using it for all transfers...

It's not 100% confirmed that the root cause is http2, but it doesn't seem device or OS specific - the same device works with arm64-v8a code, and doesn't allow receiving files on armv7a - it receives some bodies, and gets stuck on some others. Also the rest of the stack works ok (TCP, TLS, encryption, there is very little code that's specific to receiving files that's not part of http2 library), there are no issues with 64 bit code.

My hypothesis is that there is some platform-specific numbers (Int or Word) that are either overflown sooner on 32bit platform or encoded differently when they are sent to the server.

We will try to add some event logging, so may add more Information here a bit later – debugging haskell code on Android is hard.

@kazu-yamamoto please help figuring it out!

kazu-yamamoto:
@epoberezkin This might fix #63.

### DIFF
--- a/Network/HTTP2/Arch/Context.hs
+++ b/Network/HTTP2/Arch/Context.hs
@@ -83,7 +83,7 @@ data Context = Context {
   -- the connection window for sending data
   , txConnectionWindow :: TVar WindowSize
   -- window update for receiving data
-  , rxConnectionInc    :: IORef Int
+  , rxConnectionInc    :: IORef WindowSize
   , pingRate           :: Rate
   , settingsRate       :: Rate
   , emptyFrameRate     :: Rate
@@ -111,7 +111,7 @@ newContext rinfo siz =
                <*> newDynamicTableForEncoding defaultDynamicTableSize
                <*> newDynamicTableForDecoding defaultDynamicTableSize 4096
                <*> newTVarIO defaultWindowSize
-               <*> newIORef 0
+               <*> newIORef (WindowSize 0)
                <*> newRate
                <*> newRate
                <*> newRate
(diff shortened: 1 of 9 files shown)


epoberezkin:
Excellent - thank you so much - will be testing it today (or maybe tomorrow:)

epoberezkin:
Alas, it didn't solve the issue yet. It might have improved it, as on one occasion it received several http2 bodies (1-4Mb size each) via different servers, but then it got stuck anyway. It more consistently fails on smaller bodies though (256kb) - not sure if it can give a hint?

I'd keep this change, but there is something else...

Do you by any chance have some device that can run armv7a code? It might be easier to debug if you did...

kazu-yamamoto:
I understand this is a wrong fix. I will make a new PR soon.

kazu-yamamoto:
@epoberezkin If you change `Network/HTTP2/Frame/Types.hs:maxWindowSize` from 2147483647 to 1073741823 in the both sides, is this issue resolved?

epoberezkin:
It’s harder to test the servers… won’t changing the client be enough? I will try

kazu-yamamoto:
Closing 'cause this is not a correct fix.

PR #67, opened 2023-04-06
using concurrent_ for servers

kazu-yamamoto:


kazu-yamamoto:
Servers using `race_`.

When `frameReceiver` in a server generates RST_Stream, it dies.
There is no grantee that `frameSender` sends RST_Stream.
Then the connection is closed by the server side.
The client receives FIN instead of RST_Stream.

This makes the "missing" test case fragile.

Let's use `concurrently_` like clients!

PR #72, opened 2023-04-19
Generalize `Client`

edsko:
The type of `run` is

```haskell
run :: ClientConfig -> Config -> Client a -> IO a
```

which looks reasonable enough. However, `Client` is a synonym for

```haskell
type Client a = (Request -> (Response -> IO a) -> IO a) -> IO a
```

The client is given a `sendRequest` function which it can use to send requests, as long as it provides a way to process the response. That too is reasonable, but the confusing part is the use of `a` in the type of `sendRequest`: why does this have to be related to the final result of the `Client`? It seems to suggest that perhaps all of these are collected in some way, but of course that cannot be the case since (the result would have to be `IO [a]` or something).

Indeed, if we look at the code, there _is_ no relation; in this commit we therefore generalize `Client` to

```haskell
type Client a = (forall b. Request -> (Response -> IO b) -> IO b) -> IO a
```

This makes more sense: we get a callback that we can call as often as we we like; whatever type we get back when we process the response is the result of that callback; and we then use this callback to produce the final result, whatever it is.

This requires _no_ code changes in the library at all; just needs `RankNTypes` enabled in a few modules.

**Backwards compatibility**: this is strictly a generalization of the existing API, so existing code will continue to work; it just gets a more precise type, and can do more. _However_, there is one limitation, which is why I had to modify one of the tests in the test suite: where before we could have a `[Client ()]`, that is no longer possible (as it would require impredicativity). I suspect that this will not really be an issue, but I do not know for sure.

### DIFF
--- a/test/HTTP2/ServerSpec.hs
+++ b/test/HTTP2/ServerSpec.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE RankNTypes #-}
 
 module HTTP2.ServerSpec where
 
@@ -158,8 +159,18 @@ runClient allocConfig =
     runHTTP2Client s = E.bracket (allocConfig s 4096)
                                  freeSimpleConfig
                                  (\conf -> C.run cliconf conf client)
-    client sendRequest = mapConcurrently_ ($ sendRequest) clients
-    clients = [client0,client1,client2,client3,client3',client3'',client4,client5]
+
+    client :: C.Client ()
+    client sendRequest = mapConcurrently_ id $ [
+          client0   sendRequest
+        , client1   sendRequest
+        , client2   sendRequest
+        , client3   sendRequest
+        , client3'  sendRequest
+        , client3'' sendRequest
+        , client4   sendRequest
+        , client5   sendRequest
+        ]
 
 -- delay sending preface to be able to test if it is always sent first
 allocSlowPrefaceConfig :: Socket -> BufferSize -> IO Config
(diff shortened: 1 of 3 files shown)


test/HTTP2/ServerSpec.hs
```diff
@@ -158,8 +159,18 @@ runClient allocConfig =
     runHTTP2Client s = E.bracket (allocConfig s 4096)
                                  freeSimpleConfig
                                  (\conf -> C.run cliconf conf client)
-    client sendRequest = mapConcurrently_ ($ sendRequest) clients
-    clients = [client0,client1,client2,client3,client3',client3'',client4,client5]
+
+    client :: C.Client ()
+    client sendRequest = mapConcurrently_ id $ [
```
kazu-yamamoto, line 164: I think we can use `foldr1 concurrently_` instead of `mapConcurrently_ id $ `.

edsko: Sure, that would be equivalent.

edsko:
Pushed a commit that uses `foldr1 concurrently_` instead of `mapConcurrently_ id`. To avoid ambiguity errors, removed the unqualified import of `Data.ByteString.Char8` and qualified one use of `isPrefixOf` as `C8.isPrefixOf` (that `C8` qualifier was already there).

kazu-yamamoto:
Rebased and merged. Thank you for your contribution!

edsko:
@kazu-yamamoto , how does your release process work? I noticed that 4.1.4 includes #78 so I figured I'd give it a go instead of my local branch (I was going to try `http2-tls`), but when I got compilation failures, and I realized that _this_ commit is _not_ included in the release. Is that intentional?

kazu-yamamoto:
- v4.1 was created to keep the backward compatibility for `Client` but to switch to  `crypton`. Non-breaking changes are included.
- `master` will be v4.2 for this PR. I'm waiting for your another PR.

edsko:
Ok, perfect, makes total sense! Re other PR, I'm inching closer towards that; lot of stuff needed to happen on my side before I could focus on http2 again but nearly there.

PR #74, opened 2023-04-20
Introduce `forkManaged`

### LINKED ISSUE #73: Bug in `sendRequest`?
It's entirely possible I just misunderstand, but I _think_ there is a bug in [`sendRequest`](https://github.com/kazu-yamamoto/http2/blob/a41dd7415d0dcdfe4e40f989fecb91d4c254f6f0/Network/HTTP2/Client/Run.hs#L76):

```haskell
let setup = addMyId mgr
let teardown _ = deleteMyId mgr
E.bracket setup teardown $ \_ -> void $ forkIO $ do
  ..
```

(this was introduced in #47). The call to `forkIO` will return immediately, so that call to `teardown` will happen immediately.

edsko:
Fixes #73.

### DIFF
--- a/Network/HTTP2/Arch/Manager.hs
+++ b/Network/HTTP2/Arch/Manager.hs
@@ -8,12 +8,13 @@ module Network.HTTP2.Arch.Manager (
   , setAction
   , stop
   , spawnAction
-  , addMyId
+  , forkManaged
   , deleteMyId
   , timeoutKillThread
   , timeoutClose
   ) where
 
+import Control.Exception
 import Data.Foldable
 import Data.IORef
 import Data.Set (Set)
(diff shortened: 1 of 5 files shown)


edsko:
@kazu-yamamoto I was confused for a while by the use of `deleteMyId` in `worker`, since I couldn't find the corresponding `addMyId`. I think I now understand what is going on, and I tried to add a few comments to explain it. If they are wrong, let me know! :)

edsko:
I think the CI failure is actually #72 . Will fix.

kazu-yamamoto (approved):
Looks excellent to me!

kazu-yamamoto:

Issue #76, opened 2023-04-25
`RST_STREAM` with `NO_ERROR` should not throw `StreamErrorIsReceived`


edsko:
Let me try to describe the issue from first principles. First, some relevant sections of the HTTP2 specification:

[Section 6.4](https://datatracker.ietf.org/doc/html/rfc7540#section-6.4):

> The `RST_STREAM` frame (`type=0x3`) allows for immediate termination of a stream.  `RST_STREAM` is sent to request cancellation of a stream _or_ to indicate that an error condition has occurred.

(my emphasis). Then, just to be complete, [Section 7](https://datatracker.ietf.org/doc/html/rfc7540#section-7):

> NO_ERROR (0x0):  The associated condition is not a result of an error.  

And most importantly, [Section 8.1](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1):

> An HTTP response is complete after the server sends -- or the client receives -- a frame with the END_STREAM flag set (including any CONTINUATION frames needed to complete a header block).  A server can send a complete response prior to the client sending an entire request if the response does not depend on any portion of the request that has not been sent and received.  When this is true, a server MAY request that the client abort transmission of a request without error by sending a RST_STREAM with an error code of NO_ERROR after sending a complete response (i.e., a frame with the END_STREAM flag). Clients MUST NOT discard responses as a result of receiving such a RST_STREAM, though clients can always discard responses at their discretion for other reasons.

In my use case (I am writing a new gRPC client library based on `http2`), this is important, because gRPC servers make use of this option to close a stream in certain cases (some relevant tickets: https://github.com/grpc/grpc-java/issues/2162, https://trac.nginx.org/test/ticket/1792). 

However, in `http2`, when the client receives a `RST_STREAM`, [it will throw a `StreamErrorIsReceived` exception](https://github.com/kazu-yamamoto/http2/blob/3aa161e8386e4fff62251178160ff50eced84607/Network/HTTP2/Arch/Receiver.hs#L439-L443). 
I _think_ that it is not possible to catch this exception (at least, in a useful way) in client code:

* The exception is thrown in the [`frameReceiver`](https://github.com/kazu-yamamoto/http2/blob/3aa161e8386e4fff62251178160ff50eced84607/Network/HTTP2/Client/Run.hs#L32), which is a background thread spawned by `http2`
[...]

PR #77, opened 2023-05-09
Give spawned threads control over async exceptions

edsko:
@kazu-yamamoto , I'm not 100% sure if I got this right, because I don't yet have a complete picture of all the control flow in `http2`.

`http2` spawns various threads on behalf of the client code. The main idea is of this PR is to give that client code a way to handle asynchronous exceptions; specifically, to make sure that they have an exception handler of their own installed before async exceptions are enabled. 

The key change is this one:

```haskell
data OutBody = 
    .. 
    -- | Streaming body takes an unmask function, a write action and a flush action.
  | OutBodyStreaming ((forall a. IO a -> IO a) -> (Builder -> IO ()) -> IO () -> IO ())
```

### Client side

On the client side, I am fairly confident that what I have is right. I have modified `forkManaged` (from #74) to not call `unmask` directly, but instead pass it as an argument to the thread, so that _it_ can decide to unmask exceptions if and where it wants; we then just pass that on as an argument to the function in the `OutBodyStreaming`.

### Server side

On the server side, I am a bit less certain. Here threads are spawned by the manager; so I have changed the `Action` type to

```haskell
newtype Action = Action { runAction :: (forall a. IO a -> IO a) -> IO () }
```

I then propagate the `unmask` argument through, and in `response` pass it to the function inside `OutBodyStreaming`; for all other calls, I just call `unmask` directly (this is important, otherwise timeouts would not be possible).

What I am unsure about is if this limits the scope of where async exceptions are enabled too much; in particular, what about the rest of `go` in `worker`? 

### Alternative approach

So far, I have only really needed this on the client, I haven't started on the server side implementation of gRPC yet; so I'm not sure if this change is even _useful_ on the server side (mostly since those threads seem to do more than just execute user code).  I suspect that it is, but I don't have hard evidence for it currently.

The only reason I made the server modifications at this point is that the client and the server share the `OutBody` type. So an alternative approach would be to decouple these, and allow the client and the server to diverge in the type of the handler; those it's nice of course to have the consistency.

Finally, if you think that this whole thing is not a good idea and we should not give the user this control, then feel free to close the PR again; it allows me to make my async exception handling more airtight, but I don't depend on it critically.

(BTW, haven't forgotten about #76 , I will get to it soon.)

### DIFF
--- a/Network/HTTP2/Arch/Manager.hs
+++ b/Network/HTTP2/Arch/Manager.hs
@@ -3,7 +3,7 @@
 --   worker threads.
 module Network.HTTP2.Arch.Manager (
     Manager
-  , Action
+  , Action(..)
   , start
   , setAction
   , stop
@@ -29,10 +29,10 @@ import Imports
 ----------------------------------------------------------------
 
 -- | Action to be spawned by the manager.
-type Action = IO ()
+newtype Action = Action { runAction :: (forall a. IO a -> IO a) -> IO () }
 
 noAction :: Action
-noAction = return ()
+noAction = Action $ \_unmask -> return ()
 
 data Command = Stop | Spawn | Add ThreadId | Delete ThreadId
 
@@ -50,6 +50,7 @@ start timmgr = do
     void $ forkIO $ go q Set.empty ref
     return $ Manager q ref timmgr
   where
+    go :: TQueue Command -> Set ThreadId -> IORef Action -> IO ()
     go q tset0 ref = do
         x <- atomically $ readTQueue q
         case x of
@@ -62,7 +63,7 @@ start timmgr = do
       where
         next tset = do
             action <- readIORef ref
-            newtid <- forkIO action
+            newtid <- mask_ $ forkIOWithUnmask $ \unmask -> runAction action unmask
             let tset' = add newtid tset
             go q tset' ref
 
@@ -85,11 +86,11 @@ spawnAction (Manager q _ _) = atomically $ writeTQueue q Spawn
 -- This guarantees that the thread ID is added to the manager's queue before
 -- the thread starts, and is removed again when the thread terminates
 -- (normally or abnormally).
-forkManaged :: Manager -> IO () -> IO ()
+forkManaged :: Manager -> ((forall a. IO a -> IO a) -> IO ()) -> IO ()
 forkManaged mgr io =
     void $ mask_ $ forkIOWithUnmask $ \unmask -> do
       addMyId mgr
-      r <- unmask io `onException` deleteMyId mgr
+      r <- io unmask `onException` deleteMyId mgr
       deleteMyId mgr
       return r
 
@@ -130,7 +131,7 @@ kill set = traverse_ killThread set
 timeoutKillThread :: Manager -> (T.Handle -> IO ()) -> IO ()
 timeoutKillThread (Manager _ _ tmgr) action = E.bracket register T.cancel action
   where
-    register = T.registerKillThread tmgr noAction
+    register = T.registerKillThread tmgr (return ())
 
 -- | Registering closer for a resource and
 --   returning a timer refresher.
--- a/Network/HTTP2/Arch/Queue.hs
+++ b/Network/HTTP2/Arch/Queue.hs
@@ -10,7 +10,7 @@ import Network.HTTP2.Arch.Types
 {-# INLINE forkAndEnqueueWhenReady #-}
 forkAndEnqueueWhenReady :: IO () -> TQueue (Output Stream) -> Output Stream -> Manager -> IO ()
 forkAndEnqueueWhenReady wait outQ out mgr =
-    forkManaged mgr $ do
+    forkManaged mgr $ \unmask -> unmask $ do
         wait
         enqueueOutput outQ out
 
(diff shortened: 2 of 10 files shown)


kazu-yamamoto:
My impression is that the change of `OutBodyStreaming` is drastic.
If we add a field to specify exception handlers in `ClientConfig`, is your requirement satisfied?

edsko:
A global exception handler like that would need to be passed an argument in order to be able to distinguish one context from another, and then I guess that context should be passed as an argument to `OutBodyStreaming` instead of the `unmask` handler. And then it's not totally clear what the type of that context should be, so we'd need a type parameter or something? Or did you have something else in mind?

kazu-yamamoto:
Would you show examples of the usage of the first argument?
I know some examples exit in this PR but they are just called at the beginning.

kazu-yamamoto:
Perhaps, it is a good idea to add a new constructor like `OutBodyStreamingUmask`.

edsko:
> Perhaps, it is a good idea to add a new constructor like OutBodyStreamingUmask.

I'd be fine with that. I will give you an example of how I am using this, just in the middle of some other things at the moment. I will get back to you shortly :)

epoberezkin:
> If we add a field to specify exception handlers in ClientConfig, is your requirement satisfied?

it would indeed be hard to use, as ClientConfig is constructed once, and the exceptions need to be handled when the request is sent... This is somewhat over my head :), but from the user's point of view I just want to be to catch exceptions thrown in a streaming function where I send the request.

edsko:
So let me sketch the way I would want to use this. As I mentioned previously, I am working on a gRPC library, based on `http2`. The only `http2` request type I use is `requestStreaming`, with two threads: one is responsible to sending messages to the peer (this thread is started under the hood by `http2`), and one that I spawn myself for receiving message from the peer. Just like @epoberezkin , I want to catch exceptions that happen in the context of _this_ call, so that if something goes wrong, the user of the library gets detailed information what went wrong and where. 

This is tricky to get completely right, so I have written a new abstraction to help me with this; see https://github.com/well-typed/grapesy/blob/main/src/Network/GRPC/Util/Thread.hs if you'd like to see the details. It's a bit like `async`, although not quite the same; but the idea is similar: we have a `TVar` that is keeping track of the state of the thread (one complication to take care of here, unlike in `async`, is that I'm not even sure if the thread will be started at all: the thread that is started under the hood by `http2` might not be). I then keep track of two such `TVar`s in the abstraction around an ongoing gRPC call (https://github.com/well-typed/grapesy/blob/4e9de99262cc47d95fac88257eb6193d92b8f18e/src/Network/GRPC/Client/Call.hs#L60); this means that the state of the threads of the call are always inspectable; if I try to send something and the thread that is meant to handle it has died, an exception will be raised at that point; no exceptions get swallowed, and they are always propagated to where they are relevant. There is a detailed discussion of the control flow at https://github.com/well-typed/grapesy/blob/4e9de99262cc47d95fac88257eb6193d92b8f18e/src/Network/GRPC/Client/Call.hs#L115-L165 .

The way that this should ideally work is something like this:

```haskell
mask_ $ forkWithUnmask $ \unmask ->   -- this bit lives in http2
   -- this bit lives in my code :
  threadBody tvar ... $ unmask $ do 
     -- do whatever needs to happen in this request
```

This would ensure that all exceptions are disabled until we are safely instead the body of `threadBody`, which would then ensure that all exceptions are caught and stored in the relevant `TVar`, so that they are part of the `Call` state. However, without this PR, this is not possible; my current looks likes like this instead:

```haskell
forkIO $ -- this lives in http2  
   -- this bit lives in my code :
  mask $ \unmask -> 
    threadBody tvar ... $ unmask $ do 
       -- do whatever needs to happen in this request
```

Almost the same, but not quite: there is a gap between that the thread gets started by `http2` and that my infrastructure kicks in for catching exceptions; if the thread is killed in that gap (due to an async exception coming from somewhere), it will not be caught, the thread state `TVar` will not be updated, and I have lost control of the situation: the thread will be dead, I will not be able to send any more messages over the request, but I can't see what's happening and can't report a good exception to the user.

It's true that I'm being a bit pedantic here, as there is no real reason to assume that an async exception will be delivered to the thread in that gap. But I don't like solutions that work _most_ of the time; I prefer a solution that works _all_ of the time :) Having the `unmask` handler be passed as an argument makes it much easier for me to reason about my code: I know the thread will be started with async exceptions masked, so the only exceptions i need to worry about arise from calls that I'm making, which I can wrap in exception handlers of their own; and I know when I'm calling unmask, so I can make sure to call that only once an exception handler is in place.

edsko:
Obsoleted by #80.

epoberezkin:
@kazu-yamamoto am I right that this is now fixed in 4.2? I was trying to migrate to it but it seems not as straightforward with all the changed types...

kazu-yamamoto:
@epoberezkin The answer is yes.

kazu-yamamoto:
@epoberezkin Off topic but I'm sure that you are interested in https://github.com/haskellfoundation/tech-proposals/pull/57

kazu-yamamoto:
@epoberezkin See 8871654. Its test code is an example of migration.

PR #78, opened 2023-05-12
Handle `RST_STREAM`/`NO_ERROR`

### LINKED ISSUE #76: `RST_STREAM` with `NO_ERROR` should not throw `StreamErrorIsReceived`

edsko:
This closes #76 .

@kazu-yamamoto , this is two commits.

1. In the first commit, I don't change any code at all, but I add a diagram in which I document the various stream state transitions and how they are implemented. I wanted to make sure that I understood these very well before adding a new transition (the one for `RST_STREAM`). The diagram I add in this commit does _not_ yet reflect the change I want to make, but rather the status quo prior to this PR.
2. In the second commit I then make the actual change, which was now actually quite simple; there were a few ways to do it, but for consistency I now _return_ the `Closed` stream state, which is then applied in `processState`; this matches what you do for all the other cases.

As I was working on my own library, I noticed that there is a second important use case for this fix (the first being the one I outline in #76): it's not just important when the server wants to tell the client "don't send any more here, we're done", but also when the server wants to tell the client "this particular method is not actually supported at all, don't even _start_ sending me anything". Not terribly relevant for you, except to let you know that I've tested both scenarios with the fix from this PR, and they are now both working correctly.

## Related bug?

However, something that had me concerned as I was documenting the state transitions: except in the case of `RST_STREAM`, a stream is _never_ really closed on the client side (as I mention below the diagram in the code, when the `streamState` `IORef` is collected, it is typically in `HalfClosedRemote` state). This doesn't matter by itself, but I started tracking when these are then GCed, and it appears that they are not GCed at all, until the entire connection to the server is closed. 

What I would have _expected_ to see when I started documenting all the state transitions is that when the client sends their final message (meaning, the thread that gets spawned by `requestStreaming` terminates), the connection transitions to `HalfClosedLocal`, at which point it can be closed entirely and forgotten about once the server closes their end too. What do you think? Should I try to submit a PR for that, also, or have I misunderstood what the intended control flow?

(I can open a separate issue about this if that's better.)

### DIFF
--- a/Network/HTTP2/Arch/Receiver.hs
+++ b/Network/HTTP2/Arch/Receiver.hs
@@ -215,6 +215,11 @@ processState HalfClosedRemote ctx strm _streamId = do
     return False
 
 -- Transition (process5)
+processState (Closed cc) ctx strm _streamId = do
+    closed ctx strm cc
+    return False
+
+-- Transition (process6)
 processState s ctx strm _streamId = do
     -- Idle, Open Body, Closed
     setStreamState ctx strm s
@@ -452,12 +457,26 @@ stream FrameWindowUpdate header bs _ s strm = do
     increaseStreamWindowSize strm n
     return s
 
--- (Not a true transition: throws an exception)
-stream FrameRSTStream header@FrameHeader{streamId} bs ctx _ strm = do
+-- Transition (stream6)
+stream FrameRSTStream header@FrameHeader{streamId} bs ctx s strm = do
     RSTStreamFrame err <- guardIt $ decodeRSTStreamFrame header bs
     let cc = Reset err
-    closed ctx strm cc
-    E.throwIO $ StreamErrorIsReceived err streamId
+
+    -- The spec mandates (section 8.1):
+    --
+    -- > When this is true, a server MAY request that the client abort
+    -- > transmission of a request without error by sending a RST_STREAM with an
+    -- > error code of NO_ERROR after sending a complete response (i.e., a frame
+    -- > with the END_STREAM flag).
+    --
+    -- We check the first part ("after sending a complete response") by checking
+    -- the current stream state.
+    case (s, err) of
+      (HalfClosedRemote, NoError) ->
+        return (Closed cc)
+      _otherwise -> do
+        closed ctx strm cc
+        E.throwIO $ StreamErrorIsReceived err streamId
 
 -- (No state transition)
 stream FramePriority header bs _ s Stream{streamNumber} = do
--- a/Network/HTTP2/Arch/Receiver.hs
+++ b/Network/HTTP2/Arch/Receiver.hs
@@ -172,6 +172,8 @@ controlOrStream ctx@Context{..} conf@Config{..} ftyp header@FrameHeader{streamId
 ----------------------------------------------------------------
 
 processState :: StreamState -> Context -> Stream -> StreamId -> IO Bool
+
+-- Transition (process1)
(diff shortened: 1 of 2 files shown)


kazu-yamamoto (approved):
Excellent!

kazu-yamamoto:
Merged.

If a stream is not GCed, it's a bug.
Please send another PR!

PR #80, opened 2023-07-24
Introduce `OutBodyStreamingUnmask`

edsko:
This is a new attempt at providing an `unmask` callback to streaming requests, replacing #77. 

As per the discussion in #77, I've new introduced `OutBodyStreamingUnmask` alongside `OutBodyStreaming`. 

This was actually a useful change. When I first wrote #77, I didn't really understand the server part of `http2` very well yet, as I had not yet implemented the server part of my gRPC library. I now have, and understand the server aspect of `http2` much better. My suspicion in #77 that I might have gotten the control flow wrong for the server API turned out to be correct: the point of this change is that user code gets control over when exceptions are unmasked in newly spawned threads. On the client side, that newly spawned thread is created in `sendRequest` (in `Network.HTTP2.Client.Run`). On the server-side, however, this works a bit differently: `http2` spawns a new thread for each incoming request (call to `setAction` in `run`, `Network.HTTP2.Server.Run`); it then does _not_ spawn a new thread for a streaming response body. 

Therefore, this PR _only_ deals with the client side, and `OutBodyStreamingUnmask` is not actually supported at all on the server side (this PR introduces `requestStreamingUnmask`, without a corresponding `responseStreamingUnmask`). It _might_ be useful to change the `Server` type synonym so that the `server` action passed to `run` would also be given an `unmask` callback, but that has now become orthogonal to this PR.

### DIFF
--- a/Network/HTTP2/Arch/Manager.hs
+++ b/Network/HTTP2/Arch/Manager.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE RankNTypes #-}
+
 -- | A thread pool manager.
 --   The manager has responsibility to spawn and kill
 --   worker threads.
@@ -9,6 +11,7 @@ module Network.HTTP2.Arch.Manager (
   , stop
   , spawnAction
   , forkManaged
+  , forkManagedUnmask
   , deleteMyId
   , timeoutKillThread
   , timeoutClose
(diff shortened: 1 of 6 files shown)


kazu-yamamoto (approved):
Beautiful!

kazu-yamamoto:
Merged.
Thank you!

PR #79, opened 2023-07-20
Don't use generic `ThreadKilled`

edsko:
By using a specific exception debugging of client code becomes easier: when we see a thread killed using `ThreadKilled`, it could come from anywhere, but when we see a thread killed using `KilledByHttp2ThreadPoolManager`, it can only come from one place.

PR #81, opened 2023-07-25
Fix missed reference to `ThreadKilled` (from #79)

edsko:
I grepped for `ThreadKilled` and found no other references (should have done that as part of #79, apologies).


kazu-yamamoto (approved):
Sorry for the delay.
LGTM.

PR #82, opened 2023-08-25
Record reason in `KilledByHttp2ThreadPoolManager`

edsko:
This is the first patch of a series of three patches; this first patch is primarily preparatory: it records the reason a server handler is killed as part of the `KilledByHttp2ThreadPoolManager`. This primarily helps debugging, but it can also serve some other functions; more on that in the next patch.


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Merged. Thanks!

PR #88, opened 2023-09-01
Adding sockaddr

kazu-yamamoto:
My project (DNS over HTTP/2) requires to log a pair of socket addresses.

This provides my `SockAddr` and peer's `SockAddr` through `Aux` to `Server`.
`Config` has breaking changes.
If `http2-tls` is used, these changes are not visible.

Cc: @edsko @epoberezkin

### DIFF
--- a/Network/HTTP2/Arch/Config.hs
+++ b/Network/HTTP2/Arch/Config.hs
@@ -25,6 +25,10 @@ data Config = Config {
     , confReadN       :: Int -> IO ByteString
     , confPositionReadMaker :: PositionReadMaker
     , confTimeoutManager :: T.Manager
+    -- | This is copied into 'Aux', if exist, on server.
+    , confMySockAddr     :: Maybe SockAddr
+    -- | This is copied into 'Aux', if exist, on server.
+    , confPeerSockAddr   :: Maybe SockAddr
     }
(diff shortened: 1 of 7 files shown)


edsko:
Thanks for the heads up! I haven't migrated to `http2-tls` yet, still using my own integration. It's on my list of things to do :)

epoberezkin:
Found it :) Thanks for the heads up.

The problem is that by the time you have TLS the sockets are well hidden inside backend closures... Could these be Maybe, so they are optional?

We don't plan to move to http-tls as we do custom CA verification and there is another use case for a more precise control of TLS sessions... For now migrated to 4.1.4, but looking at some addressed vulnerability recently it would be great to move to the latest...

kazu-yamamoto:
If `SockAddr` is not available, please specify a dummy one (e.g. `SockAddrInet 0 0`).

epoberezkin:
yep, that's what I ended iup doing :)

Testing 4.2.2 now

epoberezkin:
It fixed the problem with the error in streaming not getting to the call! 🚀

Hopefully it'll fix some of the file sending errors we have

kazu-yamamoto:
Great!

PR #89, opened 2023-09-04
Client sync threads

### LINKED ISSUE #87: `GOAWAY` frame can overtake final `DATA` frame
(This is the corrected issue from #86 ).

I was getting non-deterministic test failures in my test suite. I think it's because when the client sends the final message to the server and then disconnects, there is a race condition between the final `DATA` frame and the `GOAWAY` frame. Successful test run:

image

(The client is sending the value "4" before disconnecting). 

Failing test run:

image
image

This is the exact same test, but the frame ordering has been reversed: since the server sees the `GOAWAY` frame _first_, it never receives the client's final message.

_Possibly_ related: #78.

I am currently using a (rather nasty) workaround by having  a `threadDelay 1_000` before closing the connection and after sending the final message.

kazu-yamamoto:
This should fix #87.

### DIFF
--- a/Network/HTTP2/Arch/Manager.hs
+++ b/Network/HTTP2/Arch/Manager.hs
@@ -16,6 +16,9 @@ module Network.HTTP2.Arch.Manager (
   , timeoutKillThread
   , timeoutClose
   , KilledByHttp2ThreadPoolManager(..)
+  , incCounter
+  , decCounter
+  , waitCounter0
   ) where
 
 import Control.Exception
@@ -41,7 +44,7 @@ noAction = return ()
 data Command = Stop (Maybe SomeException) | Spawn | Add ThreadId | Delete ThreadId
 
 -- | Manager to manage the thread pool and the timer.
-data Manager = Manager (TQueue Command) (IORef Action) T.Manager
+data Manager = Manager (TQueue Command) (IORef Action) (TVar Int) T.Manager
 
 -- | Starting a thread pool manager.
 --   Its action is initially set to 'return ()' and should be set
@@ -51,8 +54,9 @@ start :: T.Manager -> IO Manager
 start timmgr = do
     q <- newTQueueIO
     ref <- newIORef noAction
+    cnt <- newTVarIO 0
     void $ forkIO $ go q Set.empty ref
-    return $ Manager q ref timmgr
+    return $ Manager q ref cnt timmgr
   where
     go q tset0 ref = do
         x <- atomically $ readTQueue q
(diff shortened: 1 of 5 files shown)


kazu-yamamoto:
@edsko Gentle ping. Do you settle down now?

Can we release a new major version after #89 and #90 are merged?

kazu-yamamoto:
Background: `http2` blocks the release of new `http2-tls`, `http3` and `h3spec` now.

edsko:
Yes, I'm back as of Monday evening (Wednesday morning now). I should have time to look at these today or tomorrow! You'll hear from me very soon :) Thanks :)

edsko:
@kazu-yamamoto It seems that this doesn't solve the problem, when I remove my `threadDelay` workaround, I still get test failures. I will check wireshark to see if it's the same thing, will report back soon.

edsko:
Ok, I looked at the Wireshark output. If I do not use my `threadDelay` workaround, then before this PR I see the final `DATA` frame after the `GOAWAY` frame. It is true that with this PR the `GOAWAY` frame is the final one, _but_ that's because the final `DATA` frame is now gone entirely (rather than being output _before_ the `GOAWAY` frame).

kazu-yamamoto:
Uhhm.
I don't understand why the last DATA is missing.
`waitCounter0` waits until all streaming threads are done.
At this time, the last DATA should be queued.
Then GOAWAY is queued.
I have no idea at this moment.

edsko:
Ok, perhaps that's my side then. I will dig.

kazu-yamamoto:
OK. I understand.
Each stream has a TBQueue.
GOAWAY can come even when the TBQueue is not empty.
So, the sender should call `decCounter` when it receives `StreamingFinished`.

kazu-yamamoto:
I hope that [287e8f6](https://github.com/kazu-yamamoto/http2/pull/89/commits/287e8f695a6604990cd97f67f5630212e0bdc380) fixes your problem.

edsko:
@kazu-yamamoto Will, good sir, I think you got it! The problem is non-deterministic so I can't be _100%_ sure, but I have one regression test which for whatever reason seemed to trigger it quite reliably; I ran over and over again for about 10 minutes and every single run passed; then I ran my full QuickCheck-based test suite for 50 minutes and it also passed. Nice work!

kazu-yamamoto:
Merged.
Thank you for your thorough testing!

Are there any issues before the next major release?

edsko:
> Are there any issues before the next major release?

None that I am currently aware of! There might be some memory leaks, I haven't gotten to the stress testing part of my library yet, but there are, they should not affect the API.

PR #90, opened 2023-09-06
Half closed local

kazu-yamamoto:
@edsko This should take over #84.

Rather than removing `HalfClosedLocal`, this patch integrates it into `Open`.

With the first commit, a test case fails.
That's why I added `isServer`.

The second commit does this integration.
The test case passes.


edsko:
@kazu-yamamoto I will test this against my own test suite after I come back from ICFP, and I will let you know. Thanks! :-)

edsko:
@kazu-yamamoto I can confirm that this works with my test suite. I will close my own PR. Thank you!

kazu-yamamoto:
Merged. Thanks!

PR #91, opened 2023-09-15
Improving the thread manager

kazu-yamamoto:
@edsko I'm trying to use the thread manager from `http3`. This PR provides the followings. What do you think?

- Exporting the thread manager APIs.
- "Spawn" ensures to delete the thread id of newly spawned thread.
- `KilledByHttp2ThreadPoolManager` -> `KilledByHttp2ThreadManager`: thread pool means worker pool. But this manager is now also used in the client side. So, I'd like to remove "pool".


edsko:
@kazu-yamamoto All looks good to me! I just ran my test suite with this PR to make sure, and all is good. And I like the fact that `deleteMyId` is no longer exported, I think that's a good improvement.

PR #97, opened 2023-11-07
Clean stream termination

edsko:
Prior to this PR, in 

```haskell
Client.run clientConfig config $ \sendRequest -> ..
```

invoking that `sendRequest` callback can result in a `BlockedIndefinitelyOnMVar` exception ("thread blocked indefinitely in an MVar operation") if the connection has already been broken; the immediate cause of this the [call to `takeMVar`](https://github.com/kazu-yamamoto/http2/blob/592ee931b3d4b95ac811cf6a933f300f0741d949/Network/HTTP2/Client/Run.hs#L37) in `Client.Run`

```haskell
rsp <- Response <$> takeMVar (streamInput strm)
```

The key change in this PR is in the definition of `streamInput`:

```haskell
streamInput :: MVar (Either SomeException InpObj)
```

Under normal circumstances we still write a `Right inpObj` to this `MVar`, but when the stream is removed prematurely (in `remove` or in `closeAllStreams`), we write `Left` the reason why the stream was closed; this will then be thrown by the call to `sendRequest`.

@kazu-yamamoto The most important thing I am unsure about is the definition of  `closedCodeToError`, and specifically the distinction between `Finished` and `Killed`, which currently map to the same exception, which doesn't feel quite right. Do you have thoughts on this?

PS. I saw from the git commit log that you are using `fourmolu`  now, but I ran it locally and it made a ton of changes, so I guess you're using a non-standard configuration...? But I couldn't see a config file in the repo, did I miss it..?

### DIFF
--- a/Network/HTTP2/Arch/Context.hs
+++ b/Network/HTTP2/Arch/Context.hs
@@ -4,6 +4,7 @@
 
 module Network.HTTP2.Arch.Context where
 
+import Control.Exception
 import Data.IORef
 import Network.HTTP.Types (Method)
 import Network.Socket (SockAddr)
@@ -191,9 +192,12 @@ halfClosedLocal ctx stream@Stream{streamState} cc = do
 closed :: Context -> Stream -> ClosedCode -> IO ()
 closed ctx@Context{oddStreamTable, evenStreamTable} strm@Stream{streamNumber} cc = do
     if isServerInitiated streamNumber
-        then deleteEven evenStreamTable streamNumber
-        else deleteOdd oddStreamTable streamNumber
+        then deleteEven evenStreamTable streamNumber err
+        else deleteOdd oddStreamTable streamNumber err
     setStreamState ctx strm (Closed cc) -- anyway
+  where
+    err :: SomeException
+    err = toException (closedCodeToError streamNumber cc)
 
 ----------------------------------------------------------------
 -- From peer
--- a/Network/HTTP2/Arch/Receiver.hs
+++ b/Network/HTTP2/Arch/Receiver.hs
@@ -175,14 +175,14 @@ processState (Open _ (NoBody tbl@(_, reqvt))) ctx@Context{..} strm@Stream{stream
                 ProtocolError
                 streamId
                 "no body but content-length is not zero"
-    halfClosedRemote ctx strm
     tlr <- newIORef Nothing
     let inpObj = InpObj tbl (Just 0) (return "") tlr
     if isServer ctx
         then do
             let si = toServerInfo roleInfo
             atomically $ writeTQueue (inputQ si) $ Input strm inpObj
-        else putMVar streamInput inpObj
+        else putMVar streamInput $ Right inpObj
+    halfClosedRemote ctx strm
     return False
 
 -- Transition (process2)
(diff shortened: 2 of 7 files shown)


edsko:
I'm not entirely sure what to do with the test suite failures. They are failing because we are getting _different_ exceptions to the ones that the test suite currently expects; we can of course change the test suite, but perhaps you consider that not the right thing to do. Please advise :)

kazu-yamamoto:
@edsko I'm very sorry.
I forgot to add `fourmolu.yaml`.
Now I have pushed it to github.

kazu-yamamoto:
Taking about already-removed streams.
Honestly, I did not consider this case deeply so far.
So, I think your approach is the right way to go!

kazu-yamamoto:
I would be nice if you could add test cases for this.
`Internal.runIO` would help.
If more APIs (such as removing streams) are necessary, I would love to implement them.

edsko:
@kazu I haven't yet added specific tests for the problem that this PR addresses, but I _have_ fixed the test suite failures. Specifically, I made two changes:

1. In `test/HTTP2/ClientSpec.hs`, I've changed the expected exception from `StreamErrorIsReceived` to `ConnectionErrorIsReceived`. Ultimately this comes from the definition of `closedCodeToError` (also in this PR), which defines

    ````haskell
    closedCodeToError :: StreamId -> ClosedCode -> HTTP2Error
    closedCodeToError sid cc =
        case cc of
            Finished -> ConnectionIsClosed
            Killed -> ConnectionIsClosed
            Reset err -> ConnectionErrorIsReceived err sid "Connection was reset"
            ResetByMe err -> BadThingHappen err
    ````
    
    We could of course change the case for `Reset` to use `StreamErrorIsReceived` instead, but since this is closing the entire stream, `ConnectionErrorIsReceived` feels more appropriate. Also, the specific test cases where I've changed this (in `CllentSpec`) are testing for missing scheme or authority, or a mismatch between host and authority, all of which feel more like connection errors rather than stream errors to me. However, I don't feel terribly strongly about this, so if you feel the definition of `closedCodeToError` should change, let me know.

2. Somewhat of a subtle change, in `Network/HTTP2/Arch/Receiver.hs`, case `process1`, I've moved the call to `halfClosedRemote` _after_ the `putMVar`; this is necessary, because in this case the stream is empty, not en error, and we should not throw an exception in `sendRequest`.

I will try to add some test cases specifically for the deadlock that this PR addresses, probably tomorrow.

(I've now also run `fourmolu` with the config in the repo on my PR.)

kazu-yamamoto:
It's OK for me to add `ConnectionIsTimeout` for `Killed`.

kazu-yamamoto:
I'm very sorry but I will merge my branch today which conflicts with this PR.
`remove` will be `deleteEven` and `deleteOdd`.

edsko:
Ok, rebased. First comment is #99. Also introduced `ConnectionIsTimeout`. Still need to add tests specifically for the problem solved by this PR.

edsko:
@kazu-yamamoto Ok, I believe this PR is ready. The first commit now adds a test case that verifies that this deadlock cannot occur (it's a bit artificial but it will serve the purpose), and the second commit fixes the deadlock (in other words, the tests fail after the first commit but pass after the second).

kazu-yamamoto (approved):
Nice work!

PR #99, opened 2023-11-10
Run fourmolu

edsko:
This just runs `fourmolu` against `master` so that I can make sure my changes in #97 are independent.


kazu-yamamoto:
Merged.

PR #104, opened 2024-01-04
Properly close streams

edsko:
This avoids another "blocked indefinitely" in client code, following on from #97. I suspect that there are probably others, still. 

@kazu-yamamoto I'm not entirely sure _which_ exception to use when the stream is closed; if `ConnectionIsClosed` is not the right one, let me know. Also, I _think_ the first `Open`/`Open` case probably never arises in practice, but this way the code doesn't make any assumptions; but in principle I think we can just leave out this case (but I'm not 100% sure).

Incidentally, I am writing a blog post about deadlock detection in STM, which also illustrates why fixes like this are important; I'll post a link once I have a preview in case you're curious.

### DIFF
--- a/Network/HTTP2/H2/Context.hs
+++ b/Network/HTTP2/H2/Context.hs
@@ -175,7 +175,20 @@ setPeerStreamID ctx sid = writeIORef (peerStreamId ctx) sid
 
 {-# INLINE setStreamState #-}
 setStreamState :: Context -> Stream -> StreamState -> IO ()
-setStreamState _ Stream{streamState} val = writeIORef streamState val
+setStreamState _ Stream{streamState} newState = do
+    oldState <- readIORef streamState
+    case (oldState, newState) of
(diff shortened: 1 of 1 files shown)


edsko:
Incidentally, without this PR, a call to `getRequestBodyChunk` may either throw a "blocked indefinitely on STM" exception if one is lucky, or simply stall forever, depending on when GC runs exactly (and whether or not in runs), and whether or not it is therefore able to detect the fact that the STM transaction is blocked indefinitely. For this reason, without this PR my `grapesy` test suite occasionally hangs in tests with unclean client termination.

alt-romes:
@kazu-yamamoto FWIW the aforementioned @edsko's blogpost has been published at https://well-typed.com/blog/2024/01/when-blocked-indefinitely-is-not-indefinite/

kazu-yamamoto:
Yes. I tweeted about the blog post already.

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Rebased and merged.

PR #105, opened 2024-01-12
Use `String` for `Authority`

edsko:
This makes encoding to and from UTF8 (as per
https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2) the responsibility of the library rather than the client.


kazu-yamamoto (approved):
This breaks some packages but let's merge this.

PR #107, opened 2024-02-03
Deal with `RST_STREAM` in `HalfClosedLocal` state

edsko:
See detailed justification in the code itself.

@kazu-yamamoto , just FYI, I'm getting closer to a first release of `grapesy`, my new gRPC library that has motivated all of my PRs against `http2` and `http2-tls`. Its own test-suite is still not 100% happy, I don't yet know if that will yield further PRs, but `grapesy` now passes the full official gRPC interop test suite against both the Python reference and (after this PR fixing a `RST_STREAM` problem), also against the C++ reference. There are a few more references to check, I have to deal with my own test suite failures, and then I still have to do some stress testing to see if there are any memory leaks, but we are getting closer to a stable product.

### DIFF
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -557,6 +557,7 @@ stream FrameRSTStream header@FrameHeader{streamId} bs ctx s strm = do
             ConnectionErrorIsSent EnhanceYourCalm streamId "too many rst_stream"
     RSTStreamFrame err <- guardIt $ decodeRSTStreamFrame header bs
     let cc = Reset err
+    closed ctx strm cc
 
     -- HTTP2 spec, section 5.1, "Stream States":
     --
@@ -582,7 +583,6 @@ stream FrameRSTStream header@FrameHeader{streamId} bs ctx s strm = do
         (HalfClosedRemote, NoError) ->
             return (Closed cc)
         _otherwise -> do
-            closed ctx strm cc
             E.throwIO $ StreamErrorIsReceived err streamId
 
 -- (No state transition)
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -558,16 +558,27 @@ stream FrameRSTStream header@FrameHeader{streamId} bs ctx s strm = do
     RSTStreamFrame err <- guardIt $ decodeRSTStreamFrame header bs
     let cc = Reset err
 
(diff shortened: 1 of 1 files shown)


Network/HTTP2/H2/Receiver.hs
```diff
@@ -558,16 +558,27 @@ stream FrameRSTStream header@FrameHeader{streamId} bs ctx s strm = do
...
 
-    -- The spec mandates (section 8.1):
+    -- HTTP2 spec, section 5.1, "Stream States":
     --
-    -- > When this is true, a server MAY request that the client abort
-    -- > transmission of a request without error by sending a RST_STREAM with an
-    -- > error code of NO_ERROR after sending a complete response (i.e., a frame
-    -- > with the END_STREAM flag).
+    -- > A stream in the "open" state may be used by both peers to send frames
+    -- > of any type. (..) From this state, either endpoint can send a frame
```
kazu-yamamoto, line 564: What does `(..)` mean?

Network/HTTP2/H2/Receiver.hs
```diff
@@ -558,16 +558,27 @@ stream FrameRSTStream header@FrameHeader{streamId} bs ctx s strm = do
...
-    -- the current stream state.
+    -- Crucially (for the specific case we're dealing with here), it continues:
+    --
+    -- > /Either endpoint/ can send a RST_STREAM frame from this state, causing
+    -- > it to transition immediately to "closed".
+    --
+    -- (emphasis not in original). This justifies the two non-error cases,
+    -- below. (Section 8.1 of the spec is also relevant, but it is less explicit
+    -- about the /either endpoint/ part.)
     case (s, err) of
```
kazu-yamamoto, line 579: In the case of `_otherwise`, `closed` is called to delete the stream from the stream table.
I'm not sure but I guess`closed` should be called in any cases.
So, `closed` should be moved to the outside of `case`.

edsko: Makes sense. I guess I introduced that bug as part of #78, apologies. I've pushed a fix, but let me run my full test suite first before merging.

edsko: Ok, all tests pass :heavy_check_mark:

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Merged and released.

edsko:
Thanks @kazu-yamamoto !

PR #108, opened 2024-02-22
Make ping rate limit configurable

edsko:
I'm dealing with some gRPC servers (written in golang) that routine exceed the limit of 4 pings/sec (increasing the limit to 10 seems to do the trick). This PR makes the ping rate limit configurable.

### DIFF
--- a/Network/HTTP2/H2/Settings.hs
+++ b/Network/HTTP2/H2/Settings.hs
@@ -25,13 +25,15 @@ data Settings = Settings
     -- ^ SETTINGS_MAX_FRAME_SIZE
     , maxHeaderListSize :: Maybe Int
     -- ^ SETTINGS_MAX_HEADER_LIST_SIZE
+    , pingRateLimit :: Int
+    -- ^ Maximum number of pings allowed per second (CVE-2019-9512)
     }
     deriving (Eq, Show)
 
 -- | The default settings.
 --
 -- >>> baseSettings
--- Settings {headerTableSize = 4096, enablePush = True, maxConcurrentStreams = Nothing, initialWindowSize = 65535, maxFrameSize = 16384, maxHeaderListSize = Nothing}
+-- Settings {headerTableSize = 4096, enablePush = True, maxConcurrentStreams = Nothing, initialWindowSize = 65535, maxFrameSize = 16384, maxHeaderListSize = Nothing, pingRateLimit = 10}
 baseSettings :: Settings
 baseSettings =
     Settings
@@ -41,12 +43,13 @@ baseSettings =
         , initialWindowSize = defaultWindowSize -- 64K (65,535)
         , maxFrameSize = defaultPayloadLength -- 2^14 (16,384)
         , maxHeaderListSize = Nothing
+        , pingRateLimit = 10
         }
 
 -- | The default settings.
 --
 -- >>> defaultSettings
--- Settings {headerTableSize = 4096, enablePush = True, maxConcurrentStreams = Just 64, initialWindowSize = 262144, maxFrameSize = 16384, maxHeaderListSize = Nothing}
+-- Settings {headerTableSize = 4096, enablePush = True, maxConcurrentStreams = Just 64, initialWindowSize = 262144, maxFrameSize = 16384, maxHeaderListSize = Nothing, pingRateLimit = 10}
 defaultSettings :: Settings
 defaultSettings =
     baseSettings
@@ -59,7 +62,7 @@ defaultSettings =
 -- | Updating settings.
 --
 -- >>> fromSettingsList defaultSettings [(SettingsEnablePush,0),(SettingsMaxHeaderListSize,200)]
--- Settings {headerTableSize = 4096, enablePush = False, maxConcurrentStreams = Just 64, initialWindowSize = 262144, maxFrameSize = 16384, maxHeaderListSize = Just 200}
+-- Settings {headerTableSize = 4096, enablePush = False, maxConcurrentStreams = Just 64, initialWindowSize = 262144, maxFrameSize = 16384, maxHeaderListSize = Just 200, pingRateLimit = 10}
 {- FOURMOLU_DISABLE -}
 fromSettingsList :: Settings -> SettingsList -> Settings
 fromSettingsList settings kvs = foldl' update settings kvs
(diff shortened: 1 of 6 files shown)


Network/HTTP2/H2/Settings.hs
```diff
@@ -41,12 +43,13 @@ baseSettings =
         , initialWindowSize = defaultWindowSize -- 64K (65,535)
         , maxFrameSize = defaultPayloadLength -- 2^14 (16,384)
         , maxHeaderListSize = Nothing
+        , pingRateLimit = 4
```
kazu-yamamoto, line 46: If you think 4 is too small, you can increase the value.

edsko:
@kazu-yamamoto , you asked a difficult question :sweat_smile: 

> If you think 4 is too small, you can increase the value.

I didn't change the default as I wasn't sure what to change it _to_, and I wasn't comfortable suggesting something. But since you asked, I spent some time trying to see what other people do. Unfortunately, it is surprisingly difficult to find some concrete guidelines about this. Plenty of sites that tell you "one way to mitigate a ping flood attack is to impose rate limits", but what those rate limits should be, nobody seems to want to say. 

A summary of what I found:

## Alternative approaches

It seems most applications impose rate limits in a different manner than `http2` currently does. Some examples:

1. Limit on the maximum number of control frames _per connection_ (rather than per second). Examples:

    * https://avinetworks.com/docs/latest/netflix-http2-dos-advisory/ The Avi REST API introduces a parameter `max_http2_control_frames_per_connection`, which seems to have been adopted verbatim also by some other suppliers (for example, [VMWare](https://docs.vmware.com/en/VMware-NSX-Advanced-Load-Balancer/22.1/Configuration_Guide/GUID-F313D7AF-5686-4242-90A2-FF67315B567B.html)). Defaults to 1,000. 
    * `nginx` hardcodes hardcodes this limit  to a maximum of 10,000 ([patch that introduces the limit](https://forum.nginx.org/read.php?29,281804,281804#msg-281804), [location in the source code](https://github.com/nginx/nginx/blob/ef96f5835468ff8d40df29b0ddbc04ec1e5e1582/src/http/v2/ngx_http_v2.c#L2887))
    * [A10 Networks](https://support.a10networks.com/support/security_advisory/http-2-rapid-reset-cve-2023-44487) has a configurable limit here, also defaulting to 10,000.

2. Compare overall traffic to traffic on the individual streams.

    * For example, in `nginx`: "as long as total traffic is many times larger than stream traffic, we consider this to be a flood" ([patch](https://mailman.nginx.org/pipermail/nginx-devel/2021-October/014356.html)). 

3. Limit _outbound_ control packets (the _response_ to the ping, as opposed to the ping itself).

    * Some examples of this are summarized in https://github.com/dotnet/runtime/pull/52862#issuecomment-848159708

4. Impose a minimum time _between_ control packets

    * The same comment that summarizes (3) above suggests this, referring to https://github.com/dotnet/runtime/pull/52862#issuecomment-846329808 as an inspiration, with a minimum ping interval of 10 seconds.

5. Most relevant for my own needs, since I am working on a gRPC library: the [gRPC Keepalive guide](https://grpc.io/docs/guides/keepalive/) defines a server parameter `PERMIT_KEEPALIVE_TIME` which sets the "minimum allowed time between a server receiving successive ping frames without sending any data/header frame" and defaults to 5 minutes -- so this is some hybrid of (2) and (4). 

## Actual ping rate limiting

For actual _rate_ limiting, I only found two suggestions:

* Not `HTTP/2` specific, but Hillstone imposes a [rate limit on all ICMP packets](https://www.hillstonenet.com/support/5.0/en/ad_config.html), defaulting to a rather liberal 1500/second.
* The most directly relevant one I found is the [Apache traffic server](https://github.com/bryancall/trafficserver/blob/716c51c338f0c5644996d5e58a15dda1808cb472/proxy/http2/Http2ConnectionState.cc#L670-L675), which limits pings to 60 pings/minute  [by default](https://github.com/bryancall/trafficserver/blob/e52f10c0805d3f67ec7e6dacec9a33f11dd0ecb1/proxy/http2/HTTP2.cc#L801) (that is 60 per _minute_, not per second).

## Underlying cause of the many pings

I _suspect_ that the reason I am seeing more pings than the 4/second that `http2` currently allows is somehow related to the rate of _requests_ (for example -- speculating -- perhaps some logic is a bit broken and new requests can cause a new keepalive ping to be sent, irrespective of when the previous one was sent). If true, that would mean that _any_ specific value of pings/second is potentially problematic. I'd have to start digging through the golong source of the server to know for sure, but it at least feels plausible as a client "broken" in this way _would_ still work with the recommendation from the guide on rate limiting (option (5), above) -- even if it doesn't quite explain everything, at least not when compared to the guidelines (but that is no surprise in the gRPC ecosystem, unfortunately) -- it certainly doesn't seem to match [the _documented_ behaviour](https://github.com/grpc/grpc-go/blob/master/Documentation/keepalive.md) of the `golang` client :shrug: 

## Conclusion

I'm not sure what to suggest. Quite frankly, I made the ping rate limit a parameter in _my_ library too, defaulting to 10/second for now, but of course that is basically passing the bucket further down the line to the application :grin: I agree with you that sensible defaults would be better. 

Option (5) from the gRPC guidelines -- that is, imposing a minimum time between successive ping frames _without sending data/header frames_ -- does seem quite sensible, not just for gRPC but also for other applications. If you want, I can try to implement that and see if it works for me, although of course that would be a bigger change to `http2` than the current PR.

Alternatively, I could imagine some kind of pluggable "rate limiting API" which could be instantiated to _any_ of the 5 options above; I'd have to spend some time thinking about what such an API should look like.

Let me know what you'd like me to do.

kazu-yamamoto:
@edsko Thank you for this research!

Perhaps, it is easy for many programmers to measure time between packets and to count the number of a specific frame but it is hard to implement rate limiting (number of packets per second). 
It took some days for me to hit upon the idea on how to implement this rate limiting.
Yes, it looks very easy if you read the code but it would be hard to implement it without reference code.

I'm not sure it's worth exploring other approaches.
I would suggest to just increase the value from 4 to 10 with the configurable scheme.

edsko:
Ok, I've discussed this with my client and we're okay with this approach. I've force-pushed a commit that changes the default to 10 (and still keeps it configurable).

edsko:
Also changed the test suite so that the ping protection test is a bit more aggressive so that it still exceeds the limit.

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Merged.
Thank you for your contributions!

edsko:
Thanks @kazu-yamamoto !

PR #111, opened 2024-04-01
 A new type synonym 

kazu-yamamoto:
This new type synonym would make other packages more readable, which depends on `http2`.


kazu-yamamoto:
@edsko Would you review this PR?

edsko:
Makes sense to me! :)

kazu-yamamoto:
Thank you for reviewing!

Issue #113, opened 2024-04-04
http-semantics


kazu-yamamoto:
`http-semantics` should be extracted from `http2`.
The dependency should be:

```
     http-semantics
            | 
     +------+------+
     |      |      |
   http1*  http2  http3

  *probably not implemented
```

Example code under `util/` in each library:

|  | Plain | Encrypted|
|---|----|---|
|HTTP/1.1|             | `tls`: `tls-client`|
|HTTP/2| `http2`: `h2c-client`| `http2-tls`: `h2-client`|
|HTTP/3| | `http3`: `h3-client`, `quic`: `quic-client`|

kazu-yamamoto:
`http-semantics` has been released. 
During this work, I found 3 bugs of `tls`,  2 bugs of `quic` and 1 bug of `http3`.
They all have been fixed.

kazu-yamamoto:
Done.

PR #114, opened 2024-04-20
Using http-semantics

kazu-yamamoto:
This implements #113.
Backward compatibility will lost for some functions of HPACK but I guess it's a minor issue.
`warp` can be built with this branch without modification.

@edsko Could you try to use this branch? If this breaks your projects, I should provide more deprecated functions or types.


edsko:
@kazu-yamamoto I had to make a few small adjustments, but only minor, and tests pass just fine. 

Two of the changes I had to make were because there two places where I could have imported from `.Client` or `.Server` previously, but instead imported from `.Internal` because the things I needed were not client or server specific, and I did not want to make an arbitrary choice). So that's on me.

One small comment is that There is a downside to re-exporting entire modules (as in re-exporting `Semantics.Server`  from `HTTP2.Server`): the haddocks for `HTTP2.Server` will just show that the other module is re-exported, but not what is _in_ that module (until you click on it). This makes browsing the documentation  a little more cumbersome. But it's very minor.

kazu-yamamoto:
Thank you for your review.
Merged.
I agree that the haddock behavior is a little bit inconvenient.

PR #115, opened 2024-04-26
Use time-manager >= 0.1.0

FinleyMcIlwaine:
The new time-manager API removes `Handle`s immediately upon `cancel`, rather than relying on a timeout to occur. This effects http2 since servers processing many streams with a very long timeout will no longer accumulate `Handle`s, as the `Handle`s will be removed immediately when the stream is closed.

This PR bumps the time-manager dependency to include these changes and bumps the http2 version to 5.2.1.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,3 +1,8 @@
+## 5.2.1
+
+* Using time-manager v0.1.0.
+  [#115](https://github.com/kazu-yamamoto/http2/pull/115)
+
 ## 5.2.0
 
 * Using http-semantics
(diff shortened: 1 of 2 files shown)


edsko:
@kazu-yamamoto The good news is that initial testing suggests that after this PR, a `grapesy/http2` server has zero memory leaks :)

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
All relating packages are released.
Thanks!

edsko:
Thanks @kazu-yamamoto !

PR #116, opened 2024-05-25
Mark final chunk as final

### LINKED ISSUE #109: `getRequestBodyChunk` should indicate if a chunk is the last one
For streaming calls, clients have two ways of indicating end-of-input to the sever:

```
NON-EMPTY DATA FRAME 1  <not marked as end-of-stream>
NON-EMPTY DATA FRAME 2  <not marked as end-of-stream>
NON-EMPTY DATA FRAME 3  <marked as end-of-stream>
```

and

```
NON-EMPTY DATA FRAME 1  <not marked as end-of-stream>
NON-EMPTY DATA FRAME 2  <not marked as end-of-stream>
NON-EMPTY DATA FRAME 3  <not marked as end-of-stream>
EMPTY DATA FRAME 4 <marked as end-of-stream>
```

`getRequestBodyChunk` does not allow servers to differentiate between these two cases; in both cases, we will first get a non-empty bytestring indicating the last piece of data, and then the _next_ call to `getRequestBodyChunk` will return an empty bytestring. Calling `getRequestBodyChunk` _again_ to figure out after-the-fact if the previously received bytestring was the final one is not an option, because that call may block. We therefore need a generalization that doesn't just give us a bytestring, but also tells us if that bytestring was marked being the last one.

For an example of where this matters, see https://github.com/well-typed/grapesy/issues/114 .

edsko:
Closes #109.
Depends on https://github.com/kazu-yamamoto/http-semantics/pull/1.

### DIFF
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -180,7 +180,7 @@ processState (Open _ (NoBody tbl@(_, reqvt))) ctx@Context{..} strm@Stream{stream
                 streamId
                 "no body but content-length is not zero"
     tlr <- newIORef Nothing
-    let inpObj = InpObj tbl (Just 0) (return "") tlr
+    let inpObj = InpObj tbl (Just 0) (return (mempty, True)) tlr
     if isServer ctx
         then do
             let si = toServerInfo roleInfo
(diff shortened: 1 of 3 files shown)


kazu-yamamoto:
Merged.
Thanks!

edsko:
Thanks for the fast response and the Hackage releases @kazu-yamamoto !

Network/HTTP2/H2/Receiver.hs
```diff
@@ -486,7 +486,7 @@ stream
                     E.throwIO $ ConnectionErrorIsSent EnhanceYourCalm streamId "too many empty data"
             else do
                 writeIORef bodyLength len
-                atomically $ writeTQueue q $ Right body
+                atomically $ writeTQueue q $ Right (body, endOfStream)
```
edsko, line 489: here

PR #117, opened 2024-05-29
Enable containers 0.7 (ghc 9.10)

edsko:


kazu-yamamoto (changes requested):
I guess it's better to remove the upper bound in this case.

edsko:
I don't understand..?

edsko:
I mean, I can remove the upper bound if you like, or set it to something really high (otherwise Hackage will not accept the package at all), but it's not an automatic given that `http2` will continue to work with each new major version (which, by definition, can introduce breaking changes)?

kazu-yamamoto:
I have suffered from the major version up of `bytestring`.
My packages do not hit the breaking changes but I needed to modify their cabal files.
I guess that `containers` follows the same pattern.
So, I would like to simply remove the upper bound.

edsko:
Removing the upper bound altogether will cause Hackage to reject the package. Will I set it to `>= 0.6 < 1`, similar to what you do with base (`>= 4.9 < 5`)?

kazu-yamamoto:
I would like to remove the upper bound only for `containers`.
I don't think the change results in a reject from Hackage.
But you don't like this approach, `>= 0.6 < 1` is acceptable.

edsko:
Oh, maybe you are right, I just checked, and it's a warning only...? :thinking: Only if it is `base` it is an actual reject..? Ok, I learned something new today :) Will fix.

kazu-yamamoto (approved):
Now LGTM

kazu-yamamoto:
Rebased and merged.

PR #119, opened 2024-06-06
Don't wait for `push` before sending headers

edsko:
In https://github.com/kazu-yamamoto/http2/commit/a76cdf3 a change was introduced which meant that a streaming request was not enqueued until at least some data was available.  The reason (I think) was that without that delay, `outputOrEnqueueAgain` would check the queue and the streaming window, realize that the queue was empty, and then re-enqueue the `Output`, causing it to be handled out-of-order (which is problematic, because streams [must be opened in order](https://datatracker.ietf.org/doc/html/rfc7540#section-5.1.1)). 

We can't simply change `outputOrEnqueueAgain` _not_ to check the queue, because if there are no output available, we should indeed process other streams first. However, we _can_ add a special case to `outputOrEnqueueAgain` so that it _doesn't_ check either whether data is available _or_ the streaming window in the case of an `OObj` object (which will initialize the stream by sending the request headers). This is sound, because the headers do not depend on the initial data (they are fully known at this point), and the streaming window only applies to `DATA` frames. 

Once we do this, we can enqueue the request before any data is available, which can be important for some applications where the request headers must be sent even if no data is available. This PR does this in 6 steps: steps 1-4 don't change the semantics at all, but do some step-by-step refactoring to make the next two changes easier to follow; step 5 makes the change to `outputOrEnqueueAgain`, and step 6 finally changes `sendStreaming` not to wait for data to become available.

Side note: I was previously working around this by calling `push` with an empty string, but this causes problems of its own. Indeed, I hope to be following up to this PR with a second PR that makes further changes; that second PR does perhaps not _rely_ on this one, but this PR _does_ make that second PR much easier to verify, I think. That second PR is not yet ready, but submitting this one already in case I'm doing something here that is completely broken.

### DIFF
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -170,11 +170,6 @@ sendRequest ctx@Context{..} mgr scheme auth (Request req) = do
             -- the ordering of responses can be out-of-order.
             -- But for clients, the ordering must be maintained.
             -- To implement this, 'outputQStreamID' is used.
-            -- Also, for 'OutBodyStreaming', TBQ must not be empty
-            -- when its 'Output' is enqueued into 'outputQ'.
-            -- Otherwise, it would be re-enqueue because of empty
-            -- resulting in out-of-order.
-            -- To implement this, 'tbqNonEmpty' is used.
             let hdr1, hdr2 :: [Header]
                 hdr1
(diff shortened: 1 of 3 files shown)


kazu-yamamoto (approved):
Easy to follow.
Looks excellent!

edsko:
Thanks @kazu-yamamoto , I appreciate that :) The next PR is coming up soon.

PR #122, opened 2024-06-18
Update http-semantics dependency, ver bump up

FinleyMcIlwaine:
It'd be nice to have a release of http2 that includes the buffer overflow fix in http-semantics (`http-semantics >= 0.1.1`). Thanks in advance!


kazu-yamamoto:
Nice idea!
I have released a new version.
Thanks!

PR #124, opened 2024-06-21
Update http-semantics dependency, ver bump up

FinleyMcIlwaine:
Another http2 version including the [latest patch from http-semantics](https://github.com/kazu-yamamoto/http-semantics/pull/4)! Thank you in advance.


kazu-yamamoto:
Thanks!

PR #126, opened 2024-06-24
Recover rxflow on closing

FinleyMcIlwaine:
Same as #125 except we use `informWindowUpdate` in `adjustRxFlow`. This causes the appropriate window updates to go through in my test case, avoiding the window exhaustion.

Before this change, I didn't observe any window update frames in wireshark. With this change, I see window update frames with size increments of 528000 coming through every 33 requests, which makes sense for my test case given that the response bodies are 16000 bytes.

### DIFF
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -86,7 +86,7 @@ run cconf@ClientConfig{..} conf client = do
         strm <- sendRequest ctx mgr scheme authority req
         rsp <- getResponse strm
         x <- processResponse rsp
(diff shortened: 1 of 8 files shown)


kazu-yamamoto (approved):
Excellent!

kazu-yamamoto:
I noticed that WindowUpdates are not sent but guessed that is not an issue.
But from your experience, WindowUpdates are necessary.
Nice catch!
Let's merge.

kazu-yamamoto:

FinleyMcIlwaine:
Thank you!

Issue #102, opened 2023-12-18
GOAWAY shouldn't immediately close the connection


akshaymankar:
From [Section about GOAWAY in the spec](https://www.rfc-editor.org/rfc/rfc9113#section-6.8):

> The GOAWAY frame (type=0x07) is used to initiate shutdown of a connection or to signal serious error conditions. GOAWAY allows an endpoint to gracefully stop accepting new streams while still finishing processing of previously established streams. This enables administrative actions, like server maintenance.
>
> There is an inherent race condition between an endpoint starting new streams and the remote peer sending a GOAWAY frame. To deal with this case, the GOAWAY contains the stream identifier of the last peer-initiated stream that was or might be processed on the sending endpoint in this connection. For instance, if the server sends a GOAWAY frame, the identified stream is the highest-numbered stream initiated by the client.

Here the Receiver just throws an exception when it receives a `GOAWAY` and so the response can never be recieved:

https://github.com/kazu-yamamoto/http2/blob/07af719d98da6edb30ef818a93b5fce15368329b/Network/HTTP2/H2/Receiver.hs#L335-L339

wilfreddenton:
This also results in "ConnectionIsClosed" being printed to stdout constantly in my servant (warp, warp-tls) server setup.

kazu-yamamoto:
Sorry. I did not notice this issue.

Is `threadDelay` before `throwIO` good enough?

kazu-yamamoto:
@akshaymankar  @wilfreddenton Gentle ping.

wilfreddenton:
Hey so I solved my problem by adding a custom exception handler to my warp server that logs any exceptions bubbling up that far instead of printing them to stdout. So I don't have any strong recommendations in regards to how to proceed with this issue.

For me though, and this is without any contextual knowledge, it seems weird that there is a check for the existence of an error and, when it doesn't exist, this exception gets thrown. I'm assuming this is done because you want calling threads to be able to catch and handle the closing of a connection. Maybe exceptions aren't the right abstraction for this though?

akshaymankar:
Yeah, I think exception is not the right way to comminicate graceful closing of the connection. The `threadDelay` is also a bit hopeful in terms of a response being served and consumed in a given amount of time.

kazu-yamamoto:
The purpose of the exception is to break `cocurrently_` of the `async` package.

kazu-yamamoto:
I understand this issue deeply now.
[...]

Issue #127, opened 2024-07-01
A new server archtecture


kazu-yamamoto:
GHC 9.6 provides `listThreads` finally.
Watching the number of threads in `http2` using `listThreads` revealed that a lot of threads are instantly spawn due to `forkAndEnqueueWhenReady`.
One waits for `TBQ` and the other waits for stream window.

To reduce the number of threads, I would like to propose another architecture for servers.
Stop using the worker pool model.
Rather, one thread is spawn for *each stream*.
The number of threads follows `maxConcurrentStreams`.

Each designated thread feeds an output chunk to the output queue from a file or streaming data *repeatedly*.
I guess that the complicated`DynaNext` is not necessary anymore.
This should drastically simplify `Sender`.
That is, `Sender` just packs output chunks from the output queue.

Stream window should be taken care by each designated thread.
Connection window should be taken care by `Sender`.

@edsko @FinleyMcIlwaine What do you think?

kazu-yamamoto:
`DynaNext` cannot be removed because `Sender` owns `confWriteBuffer`.
Works cannot manipulate the buffer.
But I think it is still possible to reduce the number of threads.

edsko:
I think in principle this makes sense. As you already point out, having threads per stream avoids the deadlock that we ran into in https://github.com/kazu-yamamoto/network-control/pull/4 . Indeed, in `grapesy` (my gRPC library that sits on top of `http2` which has been motivating all the PRs that we've been submitting the last year), I spawn two threads for each stream, one to read incoming messages and one to send outgoing messages. 

The devil is always in the detail of course; some rethinking of the architecture might be useful, but without actually thinking it through completely it's hard for me to really say whether this will be an overall improvement or not.

That being said, I think if the goal is simplification of the model, and making the code easier to understand, I'm all for it; if the goal is merely to reduce the number of threads, then I'm less convinced that it's terribly useful; threads in Haskell are cheap enough. Indeed, the number of concurrent streams is significantly higher than the number of workers currently employed by `http2` (*).

(*) Caveat: this is actually only true for non-streaming bodies; `grapesy` exclusively uses streaming requests and responses, which means that we actually have as many workers as there are streams, because `http2` will spawn a new worker each time a streaming response is initiated.

kazu-yamamoto:
Done via #130.

PR #130, opened 2024-07-03
New server arch

kazu-yamamoto:
This implements #127

### DIFF
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -66,8 +66,8 @@ defaultClientConfig =
 -- | Running HTTP/2 client.
 run :: ClientConfig -> Config -> Client a -> IO a
 run cconf@ClientConfig{..} conf client = do
-    (ctx, mgr) <- setup cconf conf
-    runH2 conf ctx mgr $ runClient ctx mgr
+    ctx <- setup cconf conf
+    runH2 conf ctx $ runClient ctx
   where
     serverMaxStreams ctx = do
         mx <- maxConcurrentStreams <$> readIORef (peerSettings ctx)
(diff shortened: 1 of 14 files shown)


kazu-yamamoto:
Test fails sometime.
This failure should be fixed with #102.

kazu-yamamoto:
@khibino Even with this PR, `bowline` can work well.

edsko:
If I update my dependencies to use this new version of `http2`, without changing anything else, my test suite fails. I will investigate why.

edsko:
All my tests simply deadlock. I will try to figure out where and why.

kazu-yamamoto:
Thank you but I don't want you to take time to understand the deadlock at this moment.
I know the current code is unstable a bit.
I should at least resolve #102.
For this purpose, I'm now labeling all threads so that `listThreads` gives me much information.

edsko:
Ok, then I shall wait :) FWIW, it seems that the client can connect to the server just fine, and the server can respond, and the client gets the initial response headers, but then the (streaming) response body is never started, and the client is left waiting.

kazu-yamamoto:
If you run `h2c-client` and `h2c-server`, they list up the current threads with their status per second. :)

kazu-yamamoto:
@edsko It would be nice if you tested the current `main`.

edsko:
@kazu-yamamoto I just tried with 80de8dbf9e20c93dc72feab600166c9b2aa7addd , same result as with aa6f979146033629121a54964e258c3a611fed7a . My test setup is that the client does a streaming request to the server; this request arrives, and the serves initiates a streaming response. The client receives the headers, but then nothing. If I insert a call to `putStrLn` in `outBodyPush` at https://github.com/kazu-yamamoto/http2/blob/80de8dbf9e20c93dc72feab600166c9b2aa7addd/Network/HTTP2/Server/Worker.hs#L150 then I can see that print statement executed with the data that the client _should_ receive, but the client does not receive anything after the response headers. There is a Wireshark log at https://github.com/well-typed/grapesy/pull/183 where we're tracking updating to the new `http2` architecture.

kazu-yamamoto:
@edsko I cannot reproduce your issue.
If you kindly inserted several `putStrLn`s into `Sender.hs`, you could see what's happening.

PR #133, opened 2024-07-11
Make all limits configurable

FinleyMcIlwaine:
This PR makes all rate limits configurable

### DIFF
--- a/Network/HTTP2/Client.hs
+++ b/Network/HTTP2/Client.hs
@@ -64,7 +64,13 @@ module Network.HTTP2.Client (
     initialWindowSize,
     maxFrameSize,
     maxHeaderListSize,
+
+    -- ** Rate limits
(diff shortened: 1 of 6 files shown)


edsko:
@kazu-yamamoto We increased the rate limit for pings from 4 to 10 in #108, but we're still seeing some (Java) clients that are exceeding this for some reason. We're also seeing the empty data frames limit being hit. After this PR we can just change these defaults our side, so that's fine, but mentioning it here just in case you feel you want to increase the defaults in `http2` itself. We will probably increase it to 100 pings/sec our side, which feels ridiculously high but we prefer to err on the side of not rejecting valid clients versus allowing some (still quite low) load from malicious clients.

kazu-yamamoto (approved):
LGTM.
Note that I don't want to change the default values at this moment.

PR #137, opened 2024-07-19
Avoid race condition

edsko:
See #136 for discussion.

I think the test failure we were seeing there is caused by the fact that `throwTo` depends on the exception being raised to the target thread (even if it perhaps is then handling it), and that was not always succeeding (specifically, we failed to kill the "H2 streaming supporter" thread spawned in `Network.HTTP2.Client.Run.sendStreaming`). I'm not 100% sure _why_ this happens, perhaps @kazu-yamamoto knows why throwing an exception to that thread would not succeed.

Fortunately, however, we didn't quite need the strong guarantees that were provided by `kill` in #136: we don't need a guarantee that the thread has already been killed, we merely need a guarantee that the thread has been removed from the _timeout_ manager, so that we avoid the race condition explained in #136. 

This PR therefore is identical to #136 except in one place: we signal that we are done (now called `signalTimeoutsDisabled`) after we have removed all threads from the timeout manager, but before we necessarily have killed them. 

I cannot reproduce the test failure after this. Test `should always send the connection preface first` was failing quite reliably for me (though only with `+RTS -N`) before this fix. After the fix I ran it 1000 times and no failures, and I ran the full test suite 100 times, also no fails.

### DIFF
--- a/Network/HTTP2/H2/Manager.hs
+++ b/Network/HTTP2/H2/Manager.hs
@@ -16,8 +16,8 @@ module Network.HTTP2.H2.Manager (
 ) where
 
 import Data.Foldable
-import Data.Set (Set)
-import qualified Data.Set as Set
+import Data.Map (Map)
+import qualified Data.Map.Strict as Map
 import qualified System.TimeManager as T
 import UnliftIO.Concurrent
 import UnliftIO.Exception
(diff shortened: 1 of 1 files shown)


kazu-yamamoto (approved):
LGTM

PR #140, opened 2024-08-06
Avoid unnecessary empty data frames at end of stream

FinleyMcIlwaine:
If a stream terminates with an empty data frame, we will send that empty data frame even if it is succeeded by trailers that are marked end of stream. That empty data frame is unnecessary. We should only send it if the payload length is non-zero and we are not sending trailers.

I discovered this after seeing `too many empty data` errors from terminating a server handling several concurrent streams with an exception.

### DIFF
--- a/Network/HTTP2/H2/Sender.hs
+++ b/Network/HTTP2/H2/Sender.hs
@@ -282,17 +282,23 @@ frameSender
             _
             reqflush = do
                 let buf = confWriteBuffer `plusPtr` off
-                    off' = off + frameHeaderLength + datPayloadLen
                 (mtrailers, flag) <- do
                     Trailers trailers <- tlrmkr Nothing
                     if null trailers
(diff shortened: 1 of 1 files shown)


FinleyMcIlwaine:
I cannot reproduce the current test failure locally after running the test on a loop for a while, on my branch or on `main`. My guess is that it is a non-deterministic failure that is not associated with these changes.

kazu-yamamoto (approved):
Nice catch!

kazu-yamamoto:
It seems to me that your logic is correct.
So, rebased and merged.
Thanks.

FinleyMcIlwaine:
@kazu-yamamoto Thanks for that! Also, we have another PR incoming soon, so don't worry about releasing these changes just yet.

kazu-yamamoto:
The previous release was for me.
I needed to deploy the new `ServerIO` for my project.

PR #142, opened 2024-08-23
Support stream cancellation

FinleyMcIlwaine:
We now support stream cancellation via the `OutBodyIface` API's `outBodyCancel` function. Some care was taken to ensure that cancelling idle streams does not result in protocol errors. In particular, `outBodyCancel` will wait until either the headers have been sent successfully or an exception occurs. If we aren't totally sure that the headers are sent, we won't send `RST_STREAM`.

### DIFF
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -236,16 +236,48 @@ sendStreaming
     -> IO (TBQueue StreamingChunk)
 sendStreaming Context{..} strm strmbdy = do
     tbq <- newTBQueueIO 10 -- fixme: hard coding: 10
+    finishedOrCancelled <- newIORef False
     let label = "H2 streaming supporter for stream " ++ show (streamNumber strm)
     forkManagedUnmask threadManager label $ \unmask -> do
         let iface =
                 OutBodyIface
                     { outBodyUnmask = unmask
                     , outBodyPush = \b -> atomically $ writeTBQueue tbq $ StreamingBuilder b NotEndOfStream
-                    , outBodyPushFinal = \b -> atomically $ writeTBQueue tbq $ StreamingBuilder b (EndOfStream Nothing)
+                    , outBodyPushFinal = \b -> do
+                        atomicWriteIORef finishedOrCancelled True
+                        atomically $ writeTBQueue tbq $ StreamingBuilder b (EndOfStream Nothing)
+                        atomically $ writeTBQueue tbq $ StreamingFinished Nothing
                     , outBodyFlush = atomically $ writeTBQueue tbq StreamingFlush
+                    , outBodyCancel = \mErr -> do
+                        -- If we cancel a stream that hasn't yet begun (no
+                        -- headers have been exchanged yet), we could
+                        -- theoretically avoid initiating the stream at all.
+                        -- This is however difficult to do, for two reasons: the
+                        -- headers have already been enqueued, and we have
+                        -- already assigned a stream ID (which assumes that the
+                        -- server will receive those the headers). Therefore, we
+                        -- instead wait for the `streamHeadersSent`. This
+                        -- ensures that the RST_STREAM frame cannot overtake the
+                        -- headers frames.
+                        headersResult <- readMVar (streamHeadersSent strm)
+                        case headersResult of
+                          Left _ ->
+                            -- If this is an exception, it does not mean the
+                            -- headers have been sent. It means we are shutting
+                            -- down, and we might not have sent headers yet, so
+                            -- we cannot send a RST_STREAM frame
+                            return ()
+                          Right () -> do
+                            already <- atomicModifyIORef finishedOrCancelled (\x -> (True, x))
+                            if already then do
+                              return ()
+                            else
+                              atomically $ writeTBQueue tbq (StreamingCancelled mErr)
                     }
-            finished = atomically $ writeTBQueue tbq $ StreamingFinished Nothing
+            finished = do
+              already <- atomicModifyIORef finishedOrCancelled (\x -> (True, x))
+              unless already $
+                atomically $ writeTBQueue tbq $ StreamingFinished Nothing
         strmbdy iface `finally` finished
     return tbq
 
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -220,14 +220,6 @@ sendHeaderBody Config{..} ctx@Context{..} sid newstrm OutObj{..} = do
         enqueueOutputSTM outputQ out
         writeTVar outputQStreamID (sid + 2)
     forkManaged threadManager "H2 worker" $ syncWithSender ctx newstrm var sync
-  where
-    nextForStreaming
-        :: TBQueue StreamingChunk
-        -> DynaNext
-    nextForStreaming tbq =
-        let takeQ = atomically $ tryReadTBQueue tbq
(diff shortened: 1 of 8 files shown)


FinleyMcIlwaine:
@kazu-yamamoto Could you please request a review from @edsko on this as well? Thank you!

Network/HTTP2/Client/Run.hs
```diff
@@ -236,16 +236,48 @@ sendStreaming
...
     forkManagedUnmask threadManager label $ \unmask -> do
         let iface =
                 OutBodyIface
                     { outBodyUnmask = unmask
                     , outBodyPush = \b -> atomically $ writeTBQueue tbq $ StreamingBuilder b NotEndOfStream
-                    , outBodyPushFinal = \b -> atomically $ writeTBQueue tbq $ StreamingBuilder b (EndOfStream Nothing)
+                    , outBodyPushFinal = \b -> do
+                        atomicWriteIORef finishedOrCancelled True
+                        atomically $ writeTBQueue tbq $ StreamingBuilder b (EndOfStream Nothing)
+                        atomically $ writeTBQueue tbq $ StreamingFinished Nothing
```
edsko, line 249: I'm a bit worried here about atomicity; what happens if an exception is thrown to the thread that calls `outBodyPushFinal`, and we do update `finishedOrCancelled` but don't enqueue the rest? (Note that such a call would run in the context of the calling code, not in the context of a `http2` created thread.) Or indeed, enqueue the `StreamingBuilder` but not the `StreamingFinished`? Perhaps it would be better to make `finishedOrCancelled` an `TVar` rather than an `IORef`, and make this a single transaction with three statements in it.

Network/HTTP2/Client/Run.hs
```diff
@@ -236,16 +236,48 @@ sendStreaming
...
+                            -- headers have been sent. It means we are shutting
+                            -- down, and we might not have sent headers yet, so
+                            -- we cannot send a RST_STREAM frame
+                            return ()
+                          Right () -> do
+                            already <- atomicModifyIORef finishedOrCancelled (\x -> (True, x))
+                            if already then do
+                              return ()
+                            else
+                              atomically $ writeTBQueue tbq (StreamingCancelled mErr)
```
edsko, line 275: .. then this can be a single transaction also.

Network/HTTP2/Client/Run.hs
```diff
@@ -236,16 +236,48 @@ sendStreaming
...
+                        -- headers have been exchanged yet), we could
+                        -- theoretically avoid initiating the stream at all.
+                        -- This is however difficult to do, for two reasons: the
+                        -- headers have already been enqueued, and we have
+                        -- already assigned a stream ID (which assumes that the
+                        -- server will receive those the headers). Therefore, we
+                        -- instead wait for the `streamHeadersSent`. This
+                        -- ensures that the RST_STREAM frame cannot overtake the
+                        -- headers frames.
+                        headersResult <- readMVar (streamHeadersSent strm)
```
edsko, line 262: I think this `MVar` can stay an `MVar` (does not need to be part of the same transaction). If we get interrupted while we wait for the `MVar`, no harm is done.

Network/HTTP2/Client/Run.hs
```diff
@@ -236,16 +236,48 @@ sendStreaming
...
+                            if already then do
+                              return ()
+                            else
+                              atomically $ writeTBQueue tbq (StreamingCancelled mErr)
                     }
-            finished = atomically $ writeTBQueue tbq $ StreamingFinished Nothing
+            finished = do
+              already <- atomicModifyIORef finishedOrCancelled (\x -> (True, x))
+              unless already $
+                atomically $ writeTBQueue tbq $ StreamingFinished Nothing
```
edsko, line 280: .. and then a single transaction here also.

Network/HTTP2/H2/Sender.hs
```diff
@@ -200,17 +206,30 @@ frameSender
...
+                CancelNext mErr -> do
+                  -- Stream cancelled
+                  case mErr of
+                    Just err ->
+                      resetStream strm InternalError err
+                    Nothing ->
+                      resetStream strm Cancel (E.toException CancelledStream)
+                  -- It is important that we return the initial offset here,
+                  -- since we might have HEADERS frames in the buffer that
+                  -- /must/ go out before the RST_STREAM
```
edsko, line 231: Although this comment is correct, it's perhaps a bit confusing: the initial offset as opposed to what? If it's "as opposed to 0" (which was our first version here) that's wrong for more reasons than just the headers: it would also affect streams other than the one we are cancelling. 

... Actually, now that I'm trying to formulate a better reason why `off` is right here, I am unsure again about ignoring `_total` (https://github.com/kazu-yamamoto/http-semantics/pull/11#discussion_r1734029953). I need to ponder this a bit more deeply.

Network/HTTP2/Server/Worker.hs
```diff
@@ -161,11 +163,28 @@ sendStreaming Context{..} strm th strmbdy = do
...
+                            if already then
+                              return ()
+                            else
+                              atomically $ writeTBQueue tbq (StreamingCancelled mErr)
                     }
-            finished = atomically $ writeTBQueue tbq $ StreamingFinished Nothing
+            finished = do
+              already <- atomicModifyIORef finishedOrCancelled (\x -> (True, x))
+              unless already $ do
+                atomically $ writeTBQueue tbq $ StreamingFinished Nothing
```
edsko, line 187: Same comments here as in the client side (let's make these single STM transactions). It would be nice if we could somehow reduce this duplication :thinking:

Network/HTTP2/H2/Sender.hs
edsko: Ok, I've pondered this more carefully, see https://github.com/kazu-yamamoto/http-semantics/pull/11#issuecomment-2314924295 . I would change the comment above to:

> Return `off0` to discard the current DATA frame under construction. Any other frames constructed previously are left untouched: these may be frames for different streams, or may even be the HEADERS frame for the current stream (which we _must_ send before we can cancel the stream; see `streamHeadersSent`.

edsko:
This depends on https://github.com/kazu-yamamoto/http-semantics/pull/11 .

edsko (changes requested):
We realized this is not quite right. We will change the PR.

Network/HTTP2/Client/Run.hs
FinleyMcIlwaine: No longer relevant after our fix today (headers don't need to be tracked, we instead track the offset correctly after enqueueing a cancel)

edsko (changes requested):
This looks good to me now, and I think we _could_ merge this now as-is, but I think we can make a two minor improvements (as well as remove that redundant constructor, mentioned above):

1. I wonder what should happen with `outBodyPush` etc once the stream is finished. This should be considered a user bug, I think, but it might be good to detect these cases and throw an exception.
2. Let's try and avoid the redundancy between the two `outBodyIface` constructions (all the more important after we also implement (1).

edsko (commented):
LGTM! @kazu-yamamoto , this PR and https://github.com/kazu-yamamoto/http-semantics/pull/11 and now ready for your review.

kazu-yamamoto:
I have released `http2-semantics` and run CI again.

kazu-yamamoto (approved):
LGTM

- added http-semantics >= 0.2.1
- formatted with fourmolu

kazu-yamamoto:

Issue #139, opened 2024-07-26
`Prelude.undefined` called from `Network.HTTP2.Client.Run.hs:154:30`


FinleyMcIlwaine:
It seems the new exception handling behavior introduced in #137 has uncovered a previously unreachable path. I'm seeing `Prelude.undefined` errors during some of the stress tests for our gRPC library

```
test-stress: Prelude.undefined
CallStack (from HasCallStack):
  undefined, called at ./Network/HTTP2/Client/Run.hs:154:30 in http2-5.3.1-95e9324b:Network.HTTP2.Client.Run
```

edsko:
I suspect this is not directly related to #137 (perhaps it made some non-determinism happen slightly differently). We're seeing this call to `undefined` when the server dies.

edsko:
I think the problem is that the background threads are terminating normally (without an exception) when the server dies (maybe only sometimes?).

kazu-yamamoto:
The background threads (`receiver`, `sender`, etc) can finish earlier than `client`.
For instance, the server can send `GOAWAY`.
We need to decide how to treat this situation and then replace the `undefined` with proper code.
Any opinion?

edsko:
Funny that you are looking at this again, we've been looking at this also :) We think the fix isn't hard, but we hadn't had the time yet to take a look, hopefully soon (this week). PR will be coming soon!

kazu-yamamoto:
Thanks in advance.

kazu-yamamoto:
GOAWAY contains a stream ID.
This is important information for clients because streams whose stream ID is larger than this value can be resent safely in a new connection.
So, I guess that a client should received this value from the background threads.

PR #146, opened 2024-09-13
Avoid `undefined` in client

FinleyMcIlwaine:
Instead of `race`ing the background threads and the client, we run them concurrently and check which finishes first. If the background threads finished first, we `wait` for the client. If the client finished first, we `cancel` the background threads and return.

@kazu-yamamoto we also added a case in `sendGoAway` in `Network.HTTP2.H2.Receiver` so that `GoAwayIsSent` is handled the same as `ConnectionIsClosed`, which appears to more closely match the behavior of the frame sender. Also, previously if `GoAwayIsSent` happened, it would be wrapped in a `BadThingHappen` which would cause the frame sender to rethrow it. Does this change seem acceptable to you?

Also, this does not communicate the stream id from a `GOAWAY` to the client if the background threads finish first, but feel free to extend this PR to add that!

### DIFF
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -142,7 +142,7 @@ setup ClientConfig{..} conf@Config{..} = do
 
 runH2 :: Config -> Context -> IO a -> IO a
 runH2 conf ctx runClient = do
-    stopAfter mgr (clientResult <$> race runBackgroundThreads runClient) $ \res ->
+    stopAfter mgr runAll $ \res ->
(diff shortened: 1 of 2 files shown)


FinleyMcIlwaine:
This solves #139, but uncovered some other uncaught exception output coming from `tls` in our stress tests of `grapesy`. That output was eliminated with this patch to `tls` https://github.com/haskell-tls/hs-tls/pull/477

kazu-yamamoto (approved):
Looks excellent.
I will revisit your concerns when I merge my current work after merging this PR.

PR #155, opened 2024-11-23
Fix leak in H2 manager

### LINKED ISSUE #154: Memory leak in http 5.3.6
The below is a memory profile of a client using http2-5.3.5 who makes repeated connections to a server and then has a short interaction with the server on that connection before closing it again and starting again:

image

With http-5.3.6 this profile looks like this instead:

image

I will try to figure out whta's happening and submit a PR.

edsko:
See `ManagedThreads`.

Closes #154.

### DIFF
--- a/Network/HTTP2/H2/Manager.hs
+++ b/Network/HTTP2/H2/Manager.hs
@@ -19,7 +19,7 @@ import Control.Concurrent.STM
 import Control.Exception
 import qualified Control.Exception as E
 import Data.Foldable
-import Data.Map (Map)
+import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import qualified System.TimeManager as T
 
@@ -28,9 +28,14 @@ import Imports
 ----------------------------------------------------------------
 
 -- | Manager to manage the thread and the timer.
-data Manager = Manager T.Manager (TVar ManagedThreads)
+data Manager = Manager T.Manager ManagedThreads
 
-type ManagedThreads = Map ThreadId TimeoutHandle
+-- | The set of managed threads
+--
+-- This is a newtype to ensure that this is always updated strictly.
+newtype ManagedThreads = WrapManagedThreads
+    { unwrapManagedThreads :: TVar (Map ThreadId TimeoutHandle)
+    }
 
 ----------------------------------------------------------------
 
@@ -49,7 +54,7 @@ cancelTimeout ThreadWithoutTimeout = return ()
 --   by 'setAction'. This allows that the action can include
 --   the manager itself.
 start :: T.Manager -> IO Manager
-start timmgr = Manager timmgr <$> newTVarIO Map.empty
+start timmgr = Manager timmgr <$> newManagedThreads
 
 ----------------------------------------------------------------
 
@@ -70,10 +75,7 @@ stopAfter :: Manager -> IO a -> (Maybe SomeException -> IO ()) -> IO a
 stopAfter (Manager _timmgr var) action cleanup = do
     mask $ \unmask -> do
         ma <- try $ unmask action
-        m <- atomically $ do
-            m0 <- readTVar var
-            writeTVar var Map.empty
-            return m0
+        m <- atomically $ modifyManagedThreads var (\ts -> (Map.empty, ts))
         forM_ (Map.elems m) cancelTimeout
         let er = either Just (const Nothing) ma
         forM_ (Map.keys m) $ \tid ->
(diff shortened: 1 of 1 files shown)


Network/HTTP2/H2/Manager.hs
```diff
@@ -19,7 +19,7 @@ import Control.Concurrent.STM
 import Control.Exception
 import qualified Control.Exception as E
 import Data.Foldable
-import Data.Map (Map)
+import Data.Map.Strict (Map)
```
edsko, line 22: I know this change doesn't do anything, it's just for clarity.

Network/HTTP2/H2/Manager.hs
```diff
@@ -122,5 +124,29 @@ withTimeout (Manager timmgr var) action =
...
+getManagedThreads = readTVar . unwrapManagedThreads
+
+modifyManagedThreads
+    :: ManagedThreads
+    -> (Map ThreadId TimeoutHandle -> (Map ThreadId TimeoutHandle, a))
+    -> STM a
+modifyManagedThreads (WrapManagedThreads var) f = do
+    threads <- readTVar var
+    let (threads', result) = f threads
+    writeTVar var $! threads' -- strict update
```
edsko, line 145: This is the critical line.

kazu-yamamoto: I'm confused.
If this is critical, `Data.Map.Strict` is good enough.

kazu-yamamoto:
I'm not completely sure but I guess that the true source of this leak is `ThreadId`.
If so, we should use `Weak ThreadId` instead.
See: https://kazu-yamamoto.hatenablog.jp/entry/2024/11/20/160218

kazu-yamamoto:
Uhhm. `Weak` is not an instance of `Eq` and `Ord`.
So, we cannot use `Map`, sigh.

kazu-yamamoto:
We can convert `ThreadId` into `Int` via `Show`.
Then we can use `IntSet (Weak ThreadId, TimeoutHandle)`.

kazu-yamamoto:
I guess that https://github.com/kazu-yamamoto/http2/tree/improve-manager fixes this thread leak.

Network/HTTP2/H2/Manager.hs
edsko: No, that is not correct. Without the `$!`, it doesn't matter which function we use, because it is never called! It will just be a thunk build up of `insert .. delete .. insert .. delete .. insert ..`. That's why it's critical to force the map to WHNF (not NF), so that we actually call the function we use.

edsko:
> I guess that https://github.com/kazu-yamamoto/http2/tree/improve-manager fixes this thread leak.

I will try and let you know.

Network/HTTP2/H2/Manager.hs
kazu-yamamoto: We are using `Strict` and `StrictData` but `threads'` is in a tuple.
That's why?

kazu-yamamoto: The `improve-manager` branch now use `modifyTVar'`.

edsko: No, that is not the reason. `Strict` does not make function application strict, it only makes bindings strict. So if we have a module in which `Strict` is _not_ enabled (standing in for `writeTVar`)

```haskell
module A where

ignore :: a -> String
ignore _ = "Hi"
```

then in

```haskell
{-# LANGUAGE Strict, StrictData #-}

module B where

import A

ex1, ex2, ex3 :: IO ()
ex1 = putStrLn (ignore undefined)
ex2 = putStrLn (ignore $! undefined)
ex3 = let x = undefined in putStrLn (ignore x)
```

`ex1` prints "Hi" whereas `ex2` and `ex3` both throw an exception.

edsko:
I can confirm that with 5.3.8 the memory leak is gone. A _am_ seeing a lot of uncaught `TimeoutThread` exceptions in my tests, significantly more than before, but will try to work around that our side.

kazu-yamamoto:
Please don't work around it your side.
`TimeoutThread ` should be caught in the API of manager.
We should fix the API.

BTW, I have moved the thread manager from `http2` to `time-manager`.
I will fix your issues in `time-manager`.

https://github.com/kazu-yamamoto/wai/blob/check-async-exception/time-manager/System/ThreadManager.hs

kazu-yamamoto:
I see that the exception is leaking in `http3` with the `time-manager` version.
I will try to understand why.

kazu-yamamoto:
It's not leaking.
Just caught in the action passed to `forkManagedTimeout`.
I should follow the rule where asynchronous exceptions of others should be re-thrown.

edsko:
Ok, let me know if you have something that you'd like me to try out. Thanks for looking into this!

kazu-yamamoto:
My bad.
`withHandle` does not catch `TimeoutThread` since it returns `IO a`.
To catch it, the return type should be `IO (Maybe a)` where `Nothing` is returned in IO monad when timeout is fired.

I'm now considering whether or not this breaking change is acceptable.

kazu-yamamoto:
@edsko I have released several packages hoping that your problems are all gone!

edsko:
Haha, _all_ my problems gone eh? :grimacing:  Now that would be a nice [Sinterklaas](https://en.wikipedia.org/wiki/Sinterklaas) gift :grin:  I will try it out and let you know! :-)

edsko:
@kazu-yamamoto , my man , I do believe you are right! All the grapesy tests pass, including our stress tests (which check for memory leaks) with `http2-5.3.9`, `http2-tls-0.4.5`, `network-run-0.4.3` and `time-manager-0.2`, with no workarounds requirement on our side. I just had to patch `time-manager` to export the `KilledByThreadManager` exception (https://github.com/yesodweb/wai/pull/1016). I'm actually on holidays as of tomorrow (for 1.5 weeks), this is a great start to my time off :) Thank you! :)

Issue #153, opened 2024-11-19
Unexpected "Thread killed by timeout manager" in 5.3.6


edsko:
I'm seeing threads die with "Thread killed by timeout manager" in 5.3.6, but not in 5.3.5 or 5.3.4. Apologies for not noticing this sooner (when we discussed #151); I had missed it because our main test suite did not catch this error, only our stress tests do. This feels like a regression of #136 / #137, but I do not know for sure. I will dig and let you know.

edsko:
It's possible that this is related to #150.

kazu-yamamoto:
Ah, I'm sure that this is relating to ef79679.
We should not re-throw the asynchronous exception thrown by the WAI time manager.
I will check this today.

kazu-yamamoto:
FYI: https://github.com/yesodweb/wai/issues/1014

edsko:
I tried reverting ef79679 but I'm still seeing this happening.

kazu-yamamoto:
This is very strange.

- `TimeoutThread` cab be thrown if `registerKillThread` is used.
- `registerKillThread` is used only in `withTimeout`.
- `withTimeout` is used inside of `forkManaged` which ignores all exceptions.

@edsko Are you using WAI timeout manger in your client side by yourself?

kazu-yamamoto:
I pushed c8453f6 which explicitly catches `TimeoutThread` in `withTimeout` according to rule 2 in https://github.com/yesodweb/wai/issues/1014.

edsko:
We initialize the WAI time manager, because the `http2` config wants it, but don't otherwise use it. I tried with c8453f6 but no difference. I think the issue is more subtle. I will continue to dig and let you know.

edsko:
Ok, have confirmed that we can deal with this our side. I will close this issue.

kazu-yamamoto:
`withTimeout` is now using `T.withHandleKillThread` which ignores `KillThread`.
So, if @edsko's analysis is correct, this issue should not happen v5.3.7.

kazu-yamamoto:
> We previously had some logic (#136, #137)  to avoid this race condition, but I guess this logic has now been removed. 

Even with the current code, `cancel` is called before throwing `KilledByHttp2ThreadManager`.
So, I don't understand why this race happens yet.

edsko:
I think the race condition is gone as of the refactoring that uses `time-manager-0.2`.

PR #156, opened 2024-11-26
Improve manager

kazu-yamamoto:
- `forkManagedTimeout` ensures that only one asynchronous exception is thrown with a lock of `IORef Bool`. (relating to #136, #137)
- Fixing the thread leak via `Weak ThreadId` and `modifyTVar'` (relating to #154)
- `TimeoutThread` and `KilledByHttp2ThreadManager` are caught cleanly (relating to #153)

### DIFF
--- a/Network/HTTP2/H2/Manager.hs
+++ b/Network/HTTP2/H2/Manager.hs
@@ -76,6 +76,10 @@ stopAfter (Manager _timmgr var) action cleanup = do
             writeTVar var Map.empty
             return m0
         let ths = Map.elems m
+        -- Managed threads may receive 'TimeoutThread' and
+        -- 'KilledByHttp2ThreadManager'. Before throwing to
+        -- 'KilledByHttp2ThreadManager' to the tagets,
+        -- let's cancel 'TimeoutThread' to avoid race.
         forM_ (map snd ths) cancelTimeout
         let er = either Just (const Nothing) ma
         forM_ (map fst ths) $ \wtid -> do
--- a/Network/HTTP2/H2/Manager.hs
+++ b/Network/HTTP2/H2/Manager.hs
@@ -112,9 +112,9 @@ forkManagedUnmask (Manager _timmgr var) label io =
         (wtid, n) <- myWeakThradId
         -- asking to throw KilledByHttp2ThreadManager to me
         let ent = ManagedThread wtid Nothing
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
I confirmed that `modifyTVar'` uses `$!`.

```haskellmodifyTVar' :: TVar a -> (a -> a) -> STM ()
modifyTVar' var f = do
    x <- readTVar var
    writeTVar var $! f x
```

kazu-yamamoto:
I'm going to merge this PR.

Network/HTTP2/H2/Manager.hs
```diff
@@ -117,10 +143,9 @@ waitCounter0 (Manager _timmgr var) = atomically $ do
...
-    T.withHandleKillThread timmgr (return ()) $ \th -> do
-        tid <- myThreadId
-        -- overriding ThreadWithoutTimeout
-        atomically $ modifyTVar var $ Map.insert tid $ ThreadWithTimeout th
-        action th
+myWeakThradId :: IO (Weak ThreadId, Int)
+myWeakThradId = do
+    tid <- myThreadId
+    wtid <- mkWeakThreadId tid
+    let n = read (drop 9 $ show tid) -- drop "ThreadId "
```
alexbiehl, line 150: @kazu-yamamoto I am late to the party but there’s fromThreadId in base. It allows you to get a stable integer identifier of a thread without resorting to the show instance.

kazu-yamamoto: I cannot find `fromThreadId` anywhere?
Which module exports it?

edsko: https://hackage.haskell.org/package/base-4.20.0.1/docs/GHC-Conc-Sync.html#v:fromThreadId

kazu-yamamoto: `fromThreadId` is very new, sigh.

edsko: FWIW, I don't think the weak references are strictly necessary. Provided that we keep the `Map` in WHNF, and we remove entries whenever the corresponding threads die, we should not be unnecessarily retaining any threads.

kazu-yamamoto:
You are right in the current code.
`Weak ThreadId` is a kind of over-spec at this moment.

The old `System.TimeManager` did not delete an entry when its thread dies...

Issue #158, opened 2025-08-07
Streaming server is can block indefinitely on abnormal client termination


edsko:
@FinleyMcIlwaine and I spent some time debugging this, and we feel reasonably confident we have gotten to the root of the problem, though we have not yet had a chance to think about how to fix it. I hope to have time early next week to look at this, but @kazu-yamamoto if you have any thoughts meanwhile, I'd love to hear them.

Analysis of what goes wrong:

* When we spawn a new thread for a streaming response in [`sendStreaming`](https://github.com/kazu-yamamoto/http2/blob/46f8197aa9209fd247d52f29395a7a78efff55d8/Network/HTTP2/Server/Worker.hs#L167), we created a bounded queue (of size 10) for the [`StreamingChunk`](https://github.com/kazu-yamamoto/http-semantics/blob/e9f5cf2bd8ff07c514349c83e8e709b136619ef1/Network/HTTP/Semantics/FillBuf.hs#L60)
* [The various functions of the `OutBodyIface`](https://github.com/kazu-yamamoto/http2/blob/46f8197aa9209fd247d52f29395a7a78efff55d8/Network/HTTP2/H2/Stream.hs#L136-L158) essentially just write to this `TBQueue`. This means that when the queue is full, these functions will block.
* The _consumption_ of this queue (where we read from it) is a bit complicated. 
    - We give the queue as argument to [`fillStreamBodyGetNext`](https://github.com/kazu-yamamoto/http-semantics/blob/e9f5cf2bd8ff07c514349c83e8e709b136619ef1/Network/HTTP/Semantics/FillBuf.hs#L99); this function (of type `DynaNext`)  is the one that ultimately reads from the queue.
    - This `DynaNext` function is [wrapped in an `OHeader`](https://github.com/kazu-yamamoto/http2/blob/46f8197aa9209fd247d52f29395a7a78efff55d8/Network/HTTP2/Server/Worker.hs#L158), and given to `syncWithSender`, which enqueues it into the main output queue processed by `frameSender`.
[...]

PR #159, opened 2025-08-12
Timeout confReadN by using ThreadManager

kazu-yamamoto:
This should fix some parts of #158.

### DIFF
--- a/http2.cabal
+++ b/http2.cabal
@@ -138,7 +138,7 @@ executable h2c-client
         bytestring,
         http-types,
         http2,
-        network-run >= 0.3 && <0.5,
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
`warp` specifies a receiving function with time-out to `confReadN`.
So, this modification would be redundant.

kazu-yamamoto:
The algorithm found in `time-manager` was created many years ago because `timeout` could not be scaled at that time.
Now standard `timeout` can be scaled and `time-manager` makes code complicated.
So, maybe it's time to replace `time-manager` with `timeout`.
Note that `quic` uses `timeout` only.

PR #163, opened 2025-12-24
fix: restore `Config` exported constructor

blackheaven:
We have some fine grained logic which needs direct access to the constructor.


kazu-yamamoto:
The constructors are intentionally hidden for the future compatibility.
Import `Internal` modules if necessary.

blackheaven:
I have missed it, sorry, thanks for your help.

PR #164, opened 2026-03-11
Fix regression in 5.4.0: re-introduce `CFinish`

edsko:
@kazu-yamamoto I've marked this as a draft PR, because I haven't yet tested this against the full `grapesy` test suite. I will, but already submitting it as a draft PR now to ask you for some early feedback, in case you feel that the approach in this PR is wrong.

## Detailed description

In #157 quite a few changes are made to the control flow. I don't fully understand the details, but one crucial change is that the `CFinal` constructor is deleted. This has some cascading consequences:

* When the `frameReceiver` notices that the connection is closed, it no longer notifies the `frameSender` thread, which therefore does not terminate. 
* Therefore the call to `concurrently_ runReceiver runSender` by `runBackgroundThreads` in `runH2` never terminates.
* Therefore `closeAllStreams` is never called
* Therefore any server handlers that are waiting for (streaming) client input (that is, are blocked on a call to `inpObjBody`) are blocked forever, instead of being notified that the client connection is closed.

On the client side instead of a call to `concurrently_` we have a call to `race`, so that the frame sender is terminated when the frame receiver returns. It is tempting to do the same also on the server, but https://github.com/kazu-yamamoto/http2/pull/67 very explicitly does _not_ do this; I don't fully understand the details here, and in particular do not know of the discussion in that ticket is still relevant with today's architecture. Reintroducing `CFinal`, albeit used in a more limited way than it was previously seemed the safer bet.

### DIFF
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -51,8 +51,12 @@ frameReceiver ctx conf@Config{..} =
         `E.finally` atomically
             (writeTVar (receiverDone ctx) True)
   where
-    handler ConnectionIsClosed = return ()
-    handler e = E.throwIO e
+    handler :: HTTP2Error -> IO ()
+    handler e = do
+        enqueueControl (controlQ ctx) $ CFinish e
+        case e of
+            ConnectionIsClosed -> return ()
+            _otherwise -> E.throwIO e
     switch = do
         labelMe "H2 receiver"
         tid <- myThreadId
--- a/Network/HTTP2/H2/Sender.hs
+++ b/Network/HTTP2/H2/Sender.hs
@@ -126,6 +126,7 @@ frameSender
 
         -- called with off == 0
         control :: Control -> IO ()
+        control (CFinish e) = E.throwIO e
         control (CFrames ms xs) = do
             buf <- copyAll xs confWriteBuffer
             let off = buf `minusPtr` confWriteBuffer
--- a/Network/HTTP2/H2/Types.hs
+++ b/Network/HTTP2/H2/Types.hs
@@ -195,7 +195,9 @@ data Sync = Done | Cont Output
 
 ----------------------------------------------------------------
 
-data Control = CFrames (Maybe SettingsList) [ByteString]
+data Control
+    = CFinish HTTP2Error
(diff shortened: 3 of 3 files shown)


kazu-yamamoto:
First, let me explain the intention of #157. Considering TCP communication, when a user program closes a socket, it leaves the TCP termination process to the kernel and can do something immediately. Older `http2` clients had to wait for the connection to be terminated. #157 is a mechanism that lets a designated thread perform the termination process and the client thread does something immediately.

Next, I'll explain thread termination processing. Using the `async` library is just a precaution to prevent thread leaks, such as in the case of asynchronous exceptions. Recently, I've been using the *event-polling model* for normal termination processing. Relevant events such as timeouts and packet arrivals are represented in STMs, and the Sender and Receiver compose the necessary STMs and check them at the beginning of the loop.

It is in line with current policy to increase the number of `STM` variables rather than reinstating `CFinal`.

kazu-yamamoto:
There is already `receiverDone :: TVar Bool` introduced. We need to debug around this variable.

kazu-yamamoto:
There should also be an `STM` variable that tells you when the client has finished, then eliminate `race`.

edsko:
Yes, I noticed `receiveDone`. Ok, I will try to see if I can change this PR to use that instead, thanks! You'll hear from me soon (I hope) :slightly_smiling_face:

edsko:
@kazu-yamamoto Ok, thanks for pointing me in the right direction. Looking more at `receiverDone` I realized that my analysis here is anyway wrong. I will close this PR and will open a new one.

edsko:
What is confusing to me is that if I start from this PR, and then make a _single_ change to `http2`, namely remove the `enqueue` of that new `CFinish` constructor, the timeout in the `grapesy` test suite is reintroduced (so in that sense this PR does fix that problem, even if it's not quite the way we want to solve it).  I have tracked that timeout down to essentially being a timeout in the call to `readTQueue` in `readSource`. So that `CFinish` being interpreted as an exception in `frameSender` is _somehow_ relevant, but I haven't yet been able to figure out why. Unlike what I stated above, the `frameSender` _is_ terminating (it _already_ checks `receiverDone`), so it's not that. I'm a bit puzzled here; I will give this another go with fresh eyes next week.

PR #166, opened 2026-03-18
Fix regression in 5.4.0: Re-throw `ConnectionIsClosed` from `frameSender`

edsko:
@kazu-yamamoto Ok, this is a new attempt at #164 . 

The `grapesy` test suite passes with this PR (as does `http2`'s own test suite). Or course, if you feel this needs further adjustment, do let me know! Once we agree on how to solve this problem, I'll be submitting one or two further PRs also.

## Deadlock

First, my analysis of the deadlock I observed in my tests. 

* Suppose we have a server currently handling a streaming request from a client, and the client abruptly disconnects.
* The `frameReceiver` will notice the disconnect, and terminate. It explicitly swallows the `ConnectionIsClosed` exception:

https://github.com/kazu-yamamoto/http2/blob/47b38d72b4c09ab8b55ac55824ae6cadbded5ab7/Network/HTTP2/H2/Receiver.hs#L54-L55

* [`checkDone`](https://github.com/kazu-yamamoto/http2/blob/47b38d72b4c09ab8b55ac55824ae6cadbded5ab7/Network/HTTP2/H2/Sender.hs#L62) in the `frameSender` will notice that the receiver has terminated ([`done`](https://github.com/kazu-yamamoto/http2/blob/47b38d72b4c09ab8b55ac55824ae6cadbded5ab7/Network/HTTP2/H2/Sender.hs#L72)), but it won't yet terminate because not all threads have terminated: not [`gone`](https://github.com/kazu-yamamoto/http2/blob/47b38d72b4c09ab8b55ac55824ae6cadbded5ab7/Network/HTTP2/H2/Sender.hs#L70).
* However (and here is the deadlock): those threads _would_ be terminated (killed, in fact) by the [call to `stopAfter` in `runH2`](https://github.com/kazu-yamamoto/http2/blob/47b38d72b4c09ab8b55ac55824ae6cadbded5ab7/Network/HTTP2/Server/Run.hs#L135). After all, the connection is closed, so any server threads that are still running are no longer relevant.

So here's the deadlock: `frameSender` is not terminating because not all server threads have terminated, but the server threads cannot terminate because they are are waiting for more input from the client and are not informed that the connection has been closed.

## Solution

I am not completely sure what the motivation is for swallowing the `ConnectionIsClosed` in `frameReceiver`, but I _guess_ it's to allow the `frameSender` to terminate cleanly. I've now generalized the `receiverDone` `TVar` to `receiverStatus` instead: when the frame receiver terminates, it records _how_: cleanly, with an exception, or because the connection is closed. Then in `frameSender` when we check the `receiverStatus`, _if_ the connection is closed we throw the `ConnectionIsClosed` exception at that point.

It is important that we throw this exception (rather than simply terminate the `frameSender`), because if we simply terminate, then to a server handler this looks like the client cleanly terminated the stream rather than got disconnected. I think this may be important within `http2` itself also; for example, `closureServer` [explicitly checks for `ConnectionIsClosed`](https://github.com/kazu-yamamoto/http2/blob/main/Network/HTTP2/H2/Receiver.hs#L656); if we don't rethrow that exception, that case would never trigger.

### DIFF
--- a/Network/HTTP2/H2/Sender.hs
+++ b/Network/HTTP2/H2/Sender.hs
@@ -67,11 +67,15 @@ checkDone Context{..} 0 = atomically $ do
         then
             return False
         else do
-            gone <- isAllGone threadManager
-            unless gone retry
-            done <- readTVar receiverDone
-            unless done retry
-            return True
+            threadMgrGone <- isAllGone threadManager
+            recvStatus <- readTVar receiverStatus
+            case (threadMgrGone, recvStatus) of
+                (_, ReceiverConnectionClosed) ->
+                    throwSTM ConnectionIsClosed
+                (True, ReceiverDone _) ->
+                    return True
+                _otherwise ->
+                    retry
 checkDone _ _ = return False
 
 frameSender :: Context -> Config -> IO ()
--- a/Network/HTTP2/H2/Context.hs
+++ b/Network/HTTP2/H2/Context.hs
@@ -90,10 +90,15 @@ data Context = Context
     , mySockAddr         :: SockAddr
     , peerSockAddr       :: SockAddr
     , threadManager      :: T.ThreadManager
-    , receiverDone       :: TVar Bool
+    , receiverStatus     :: TVar ReceiverStatus
     , workersDone        :: STM Bool
     }
 {- FOURMOLU_ENABLE -}
+
+data ReceiverStatus
+    = ReceiverRunning
+    | ReceiverDone (Maybe SomeException)
+    | ReceiverConnectionClosed
 
 ----------------------------------------------------------------
 
@@ -138,7 +143,7 @@ newContext roleInfo Config{..} cacheSiz connRxWS mySettings timmgr mdone = do
     let mySockAddr   = confMySockAddr
     let peerSockAddr = confPeerSockAddr
     threadManager   <- T.newThreadManager timmgr
-    receiverDone    <- newTVarIO False
+    receiverStatus  <- newTVarIO ReceiverRunning
     let workersDone = fromMaybe (T.isAllGone threadManager) mdone
     return Context{..}
   where
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -46,13 +46,20 @@ headerFragmentLimit = 51200 -- 50K
 ----------------------------------------------------------------
 
 frameReceiver :: Context -> Config -> IO ()
-frameReceiver ctx conf@Config{..} =
-    (switch `E.catch` handler)
-        `E.finally` atomically
-            (writeTVar (receiverDone ctx) True)
+frameReceiver ctx@Context{receiverStatus} conf@Config{..} =
+    E.mask $ \unmask -> do
+        mErr <- E.try $ unmask switch
+        case mErr of
+            Left err | Just ConnectionIsClosed <- E.fromException err -> do
+                atomically $ writeTVar receiverStatus $ ReceiverConnectionClosed
+                return ()
+            Left err -> do
+                atomically $ writeTVar receiverStatus $ ReceiverDone (Just err)
+                E.throwIO err
+            Right () -> do
+                atomically $ writeTVar receiverStatus $ ReceiverDone Nothing
+                return ()
   where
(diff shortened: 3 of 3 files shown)


edsko:
Hmmm @kazu-yamamoto I'm noticing some non-deterministic behaviour in `grapesy`. Not yet sure if the fix in this PR is insufficient. Either way would be good to hear if you think the approach in this PR makes sense or not.

edsko:
I'll mark this as a draft until I know why we have a non-deterministic failure in our test suite.

edsko:
Ok, I tracked the non-determinism down to a bug in our test suite, which had apparently been present for a long time and was just much more likely to be triggered now. So I _think_ this PR is good, but I will do some more due diligence ours side first before really calling this ready.

Network/HTTP2/H2/Sender.hs
```diff
@@ -67,11 +67,15 @@ checkDone Context{..} 0 = atomically $ do
         then
             return False
         else do
-            gone <- isAllGone threadManager
-            unless gone retry
-            done <- readTVar receiverDone
-            unless done retry
-            return True
+            threadMgrGone <- isAllGone threadManager
```
kazu-yamamoto, line 70: I guess that `gone` is a better name than `threadMgrGone` because `threadMgrGone` sounds like only the manager is gone.

Network/HTTP2/H2/Sender.hs
```diff
@@ -67,11 +67,15 @@ checkDone Context{..} 0 = atomically $ do
...
+            recvStatus <- readTVar receiverStatus
+            case (threadMgrGone, recvStatus) of
+                (_, ReceiverConnectionClosed) ->
+                    throwSTM ConnectionIsClosed
```
kazu-yamamoto, line 74: Why throwing `ConnectionIsClosed` here?
What about `return True`?

Network/HTTP2/H2/Sender.hs
edsko: Ah, fair enough. Yes, makes sense, will change.

Network/HTTP2/H2/Sender.hs
edsko: This is what I tried to explain in the PR description:

It is important that we throw this exception (rather than simply terminate the frameSender), because if we simply terminate, then to a server handler this looks like the client cleanly terminated the stream rather than got disconnected. I think this may be important within http2 itself also; for example, closureServer [explicitly checks for ConnectionIsClosed](https://github.com/kazu-yamamoto/http2/blob/main/Network/HTTP2/H2/Receiver.hs#L656); if we don't rethrow that exception, that case would never trigger.

If you prefer we return normally, then we'd have to deal with these two use cases some other way?

kazu-yamamoto: I understand the deadlock.

Before merging this PR, could I ask for one more experiment?

Is it possible to use `stopAfter` in `runReceiver` to make it independent of the Sender and avoid relying on the `ConnectionIsClosed` exception?

kazu-yamamoto: My thought is here:

```haskell
runH2 :: Config -> Context -> IO ()
runH2 conf ctx = do
    let mgr = threadManager ctx
        runReceiver = T.stopAfter mgr (frameReceiver ctx conf) $ \res ->
            closeAllStreams (oddStreamTable ctx) (evenStreamTable ctx) res
        runSender = frameSender ctx conf
    er <- E.try $ concurrently_ runReceiver runSender
    case er of
        Right () -> return ()
        Left e -> closureServer conf ctx e
```

kazu-yamamoto: You may add `ReceiverStatus` if you think necessary.

edsko: I think it's possible in principle, but it worries me slightly to close all streams with the sender still running. After all, especially the even streams might be used by the frameSender? 

But perhaps a version of this PR that doesn't rethrow the exception would clarify things. I'll try that, and then you can see what you prefer, or if indeed you really want to close all streams with the sender potentially still running. You'll hear from me soon (I hope).

edsko:
Superseded by #167.

PR #167, opened 2026-03-28
Ensure sender notices when receiver has terminated

edsko:
@kazu-yamamoto This is an alternative to #166, which does not depend on any exceptions being re-thrown. Instead of having 

```hs
receiverDone :: TVar Bool
```

in the context, we now have

```hs
receiverDone :: TVar (Maybe SomeException)
```

The receiver only ever terminates with an exception, which this makes clear. The sender then reads this `TVar`, and if it notices that the receiver has thrown an exception, it terminates with that same exception. Both the sender and the receiver also return this exception.

Since the exception is now returned as a value, we need a slight generalization of `stopAfter`, see https://github.com/yesodweb/wai/pull/1069 .

### DIFF
--- a/Network/HTTP2/Client/Run.hs
+++ b/Network/HTTP2/Client/Run.hs
@@ -151,8 +151,15 @@ runH2 conf ctx runClient = do
         er <- race runReceiver runClient
         case er of
             Right r -> return r
-            -- never reached because runReceiver throws an exception to exit.
-            Left () -> throwIO ConnectionIsClosed
+            Left err -> throwIO err
+
+    -- When 'runClientReceiver' terminates, it is important we give the sender
+    -- a chance to terminate cleanly also (it's possible the client terminated
+    -- but there are still some messages in the queue to be sent).
+    --
+    -- If the client terminated successfully, we ignore any other errors in the
+    -- sender (indeed, any exception here might simply be that the background
+    -- threads were cancelled /because/ the client terminated).
     runAll = snd <$> concurrently runSender runClientReceiver
 
 makeStream
--- a/Network/HTTP2/H2/Context.hs
+++ b/Network/HTTP2/H2/Context.hs
@@ -90,7 +90,7 @@ data Context = Context
     , mySockAddr         :: SockAddr
     , peerSockAddr       :: SockAddr
     , threadManager      :: T.ThreadManager
-    , receiverDone       :: TVar Bool
+    , receiverDone       :: TVar (Maybe SomeException)
     , workersDone        :: STM Bool
     }
 {- FOURMOLU_ENABLE -}
@@ -138,7 +138,7 @@ newContext roleInfo Config{..} cacheSiz connRxWS mySettings timmgr mdone = do
     let mySockAddr   = confMySockAddr
     let peerSockAddr = confPeerSockAddr
     threadManager   <- T.newThreadManager timmgr
-    receiverDone    <- newTVarIO False
+    receiverDone    <- newTVarIO Nothing
     let workersDone = fromMaybe (T.isAllGone threadManager) mdone
     return Context{..}
   where
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -19,6 +19,7 @@ import qualified Data.ByteString.Char8 as C8
 import qualified Data.ByteString.Short as Short
 import qualified Data.ByteString.UTF8 as UTF8
 import Data.IORef
+import Data.Void
 import Network.Control
 import Network.HTTP.Semantics
 import qualified System.ThreadManager as T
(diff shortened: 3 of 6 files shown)


edsko:
I am seeing some non-deterministic behaviour stilli in the grapesy test suite, but I confirmed that I see this also with 5.3.9. I do not yet know if it's a problem in http2 or in grapesy. I'll mark this PR as a draft therefore, even though I think this is entirely unrelated.

kazu-yamamoto:
What is the current status of this issue?

edsko:
@kazu-yamamoto Thanks for asking! This PR depends on [`stopAfterWithResult`](https://hackage-content.haskell.org/package/time-manager-0.3.2/docs/System-ThreadManager.html#v:stopAfterWithResult), which is now released. However, there is some flakiness in the `grapesy` test suite that I had not noticed before. I now know that this is due to an upgrade of `tasty`: the latest version of `tasty` sets the RTS flag `-N` at runtime. The `grapesy` test suite prior to that was running without any `-N` RTS flag (though it was using the `-threaded` runtime), and as a result some of the tests made certain assumptions about ordering that are not justified now. I want to fix those first, before I can test this PR more thoroughly. I _think_ this PR is in principle good to go, but until I can _really_ test it properly with my full `grapesy` test suite, I feel I lack evidence to _really_ state with confidence that it's right.

So if you want to merge it already, I would not necessarily object, but my personal preference would be to keep it open a bit longer until I can really give the green light. All the time I have at the moment for actual coding goes to `grapesy` (and, by implication, `http2`), so it's my current focus, but debugging non-deterministic failures can be pretty time consuming, as I am sure you are well aware :grin: I will for sure keep you up to date! (And I anyway I have some other patches against `http2` that I want to submit PRs for, but they are currently based off 5.3.9 rather than 5.4.0, and I want to resolve these issues first before "forward porting" those other changes).

Hope that's all ok!

kazu-yamamoto:
I respect your request and will wait. My DNS cache server (bowline) also has an issue where workers freeze when using DNS over HTTP/2. It's difficult to reproduce, so I can't say for sure, but I think this PR might solve it.

edsko:
Apologies for the delay @kazu-yamamoto . I've not been well. I'll continue to look at this next week and will keep you posted.

kazu-yamamoto:
@edsko Take care.

edsko:
@kazu-yamamoto I'm finally back in a position to look at this PR carefully, sorry it took so long. Combination of my health and struggling hugely to get proper exception annotations ([blogpost](https://well-typed.com/blog/2026/05/lay-annotation-land/)). But I'm finally back at a stage now where I can test this PR properly, there is no more flakiness in the `grapesy` test suite as far as I am aware. _However_, our "abrupt disconnect" tests are failing with this PR (they pass with http2-5.3.9), so it seems this PR isn't _quite_ right yet. This is now at the top of my priority list, so hopefully I should have some answers for you soon(ish).

edsko:
The good news is that I have a reliable reproducer for the problem, so hopefully I should be able to say with some certainty once I fix it that the deadlock _is_ indeed fixed.

kazu-yamamoto:
I hope you are doing well.
Thank you for taking the time to address this issue.

edsko:
I have pushed a commit which I _think_ fixes this PR; see commit message for details. However, with this commit in place I do see some non-deterministic failures in the grapesy test-suite (though the reliable reproducer that I was using to check for this deadlock is now passing every time). I don't yet know if that's a grapesy problem or a http2 problem, I will report back.

edsko:
@kazu-yamamoto Ok, we're still not there I'm afraid. Still getting non-deterministic failures in the grapesy test suite. I'm also not yet sure about that last commit I pushed, I might have to undo that again. 

It is _so_ difficult to get this air-tight :pensive: Unfortunately I am out of time for now. I'm going to have to come back to this, and I will, but it will probably be in a week or so.

kazu-yamamoto:
@edsko Thank you for your effort. Of course, I can wait.

edsko:
@kazu-yamamoto My good sir, I can finally confirm that this PR is good to go. The last non-determinism that I was seeing was on the side of `grapesy`. 

The third commit I had previously pushed to this PR was also correct; it's just that the _justification_ of the change wasn't correct. For your ease of review, I've merged the three commits now, because commit 2 was doubling back on commit 1 a bit. Compared to the PR as you saw it, the only new change is in `closeAllStreams`, and there the only change I've made is to add a comment that explains _why_ the change was made.

I will have have some follow-up PRs to do with cancellation, that were blocked on my figuring out was was going on here; but given how long this one took, I'd not like to make any predictions. Hopefully, it won't take as long, but equally it would make sense to release this as `http2-5.4.1` if you wanted to. The next fix that I'm going to propose will need to touch `http-semantics` as well (albeit only in the comments), so it's somewhat broader in scope.

Apologies again for the very slow turn-around, this was a tough one.

kazu-yamamoto (approved):
LGTM.

kazu-yamamoto:
Version 5.4.1 has been released.
Thank you for your contributions!

PR #168, opened 2026-07-09
Support sending informational (1xx) responses like 103 Early Hints

thomasjm:
This is the second step of a plan I'm working on to support HTTP early hints in WAI/Warp and associated libraries.

We initially talked about this in https://github.com/yesodweb/wai/issues/970 and now I'm hoping to do it properly.

Depends on https://github.com/kazu-yamamoto/http-semantics/pull/18

### DIFF
--- a/Network/HTTP2/H2/Receiver.hs
+++ b/Network/HTTP2/H2/Receiver.hs
@@ -387,6 +387,27 @@ checkPriority p me
   where
     dep = streamDependency p
 
+-- | Handle a decoded response HEADERS section. On the client, a 1xx
+--   informational response (e.g. 103 Early Hints) is delivered to the
+--   informational callback and the stream keeps waiting for the final response;
+--   otherwise the headers become the (final) response.
+onResponseHeaders
+    :: Context
+    -> StreamId
+    -> Maybe ClosedCode
+    -> Bool
+    -> TokenHeaderTable
+    -> IO StreamState
+onResponseHeaders ctx streamId hcl endOfStream tbl
+    | endOfStream = return $ Open hcl (NoBody tbl)
+    | role ctx == Client && isInformational = do
+        informationalCallback ctx streamId tbl
+        return $ Open hcl JustOpened
+    | otherwise = return $ Open hcl (HasBody tbl)
+  where
+    isInformational =
+        maybe False ("1" `BS.isPrefixOf`) $ getFieldValue tokenStatus (snd tbl)
+
 stream
     :: FrameType
     -> FrameHeader
@@ -416,11 +437,7 @@ stream FrameHeaders header@FrameHeader{flags, streamId} bs ctx s@(Open hcl JustO
             if endOfHeader
                 then do
                     tbl <- hpackDecodeHeader frag streamId ctx
-                    return $
-                        if endOfStream
-                            then -- turned into HalfClosedRemote in processState
-                                Open hcl (NoBody tbl)
-                            else Open hcl (HasBody tbl)
+                    onResponseHeaders ctx streamId hcl endOfStream tbl
                 else do
                     let siz = BS.length frag
                     return $ Open hcl $ Continued [frag] siz 1 endOfStream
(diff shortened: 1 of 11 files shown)


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Merged.

Network/HTTP2/Server/Worker.hs
```diff
@@ -29,6 +30,7 @@ runServer conf server ctx@Context{..} strm req =
                     { auxTimeHandle = th
                     , auxMySockAddr = mySockAddr
                     , auxPeerSockAddr = peerSockAddr
+                    , auxSendInformational = sendInformational ctx strm
```
Vlix, line 33: @thomasjm @kazu-yamamoto This will cause build errors because the lower bounds of `http-semantics` hasn't been changed.
Please revise the current `http2-5.4.2` on hackage to `http-semantics >= 0.4.1`, and update the `.cabal` file accordingly.

---

@kazu-yamamoto I feel like more care could be taken to preserve backwards compatibility.
Down the line in the `wai` repo, we constantly have to fix CI by hard-coding version numbers because dependencies keep breaking backwards compatibility.

This could have been wrapped in CPP so that `< 0.4.1` would also be supported

```hs
#if MIN_VERSION_http_semantics(0,4,1)
  , auxSendInformational = sendInformational ctx strm
#endif

-- and the definition of `sendInformational` as well, of course
```

Network/HTTP2/H2/Receiver.hs
```diff
@@ -387,6 +387,27 @@ checkPriority p me
...
+onResponseHeaders
+    :: Context
+    -> StreamId
+    -> Maybe ClosedCode
+    -> Bool
+    -> TokenHeaderTable
+    -> IO StreamState
+onResponseHeaders ctx streamId hcl endOfStream tbl
+    | endOfStream = return $ Open hcl (NoBody tbl)
+    | role ctx == Client && isInformational = do
```
Vlix, line 403: Will this not break other `1XX` responses? Like `100`?
Because when getting `100`, the receiver shouldn't wait for more, it should send the actual request.
`101` is not a thing in HTTP/2, IIRC, but still, if someone uses a custom `1XX` response, will this break? Because it will keep waiting for more?

What I'm saying is: it feels very "103 specific", but it's implemented as `1XX`.

Vlix (commented):
I think the CI of this repo doesn't catch backwards compatibility breakage, since it only uses regular cabal which will _always_ use the newest versions.

It might be a good idea to also have one runner that runs a stackage LTS with the lowest `base` version that is supported, to see that the earliest versions supported still build.

Network/HTTP2/Server/Worker.hs
thomasjm: I'd be happy to do a follow-up to wrap it in CPP if you all would like.

Network/HTTP2/H2/Receiver.hs
thomasjm: Hmm, it's true that this code formerly left the stream as `Open hcl (HasBody tbl)` after a `100`, but now it returns `Open hcl JustOpened`.

I looked into this a little bit and it seems like this is actually an improvement across the board. AFAICT, `JustOpened` is the correct state to stay in after any `1xx` response. Because if you go into `HasBody`, then it expects the next frame to be a body, and you can't send any more `1xx` blocks.

The original behavior would preclude something like the following:

1. Client sends `Expect: 100-continue`
2. Server responds with `100`
3. Client sends request body
4. Server sends a `103` with some early hints, followed by a `200` with a response body.

Network/HTTP2/Server/Worker.hs
kazu-yamamoto: Let's go with CPP.
Deprecate `http2` 5.4.2 and release 5.4.3.

thomasjm: https://github.com/kazu-yamamoto/http2/pull/170

thomasjm: Looks like 5.4.2 is deprecated and 5.4.3 is released now; resolving this

PR #172, opened 2026-08-13
Improvements for dealing with RST_STREAM

edsko:
Hi @kazu-yamamoto ,

This is a series of commits, each of which fixes a different aspect of `RST_STREAM` ; I've tried to document each commit separately, let me know if anything is unclear. Two things I'd particularly like your eye on, both to do with `H2.Sync`:

* I modified `checkLoop` to check that the stream is still open, preventing a wait for a window update that will never come. This _might_ be a regression (#107)
* For `cancelAfterFinish`, I'm using `makeOutputIO`, which does no actual syncing; I think this is fine, since it's only sending a `RST_STREAM`, which is not subject to window size restrictions anyway, but I wasn't 100% sure.

I ran the full grapesy test suite and gRPC interop tests against this patch; everything green my side.

Heads up: we _may_ also have found a HPACK bug which can happen when `RST_STREAM` is received, but I'm not entirely sure yet. I don't have a reproducer yet; it's anyway unrelated to _this_ PR.

### DIFF
--- a/Network/HTTP2/H2/OutBodyIface.hs
+++ b/Network/HTTP2/H2/OutBodyIface.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE RankNTypes #-}
 
 module Network.HTTP2.H2.OutBodyIface (
+    StreamTerminated (..),
     withOutBodyIface,
 ) where
 
@@ -19,6 +20,7 @@ data StreamTerminated
     = StreamPushedFinal
     | StreamCancelled
     | StreamOutOfScope
+    | StreamRemoteReset ClosedCode
     deriving (Show)
     deriving anyclass (Exception)
 
@@ -31,12 +33,39 @@ withOutBodyIface
     -> (forall a. IO a -> IO a)
     -> (OutBodyIface -> IO r)
     -> IO r
-withOutBodyIface _ctx _strm tbq unmask k = do
+withOutBodyIface _ctx strm tbq unmask k = do
     terminated <- newTVarIO Nothing
     let checkNotTerminated :: STM ()
         checkNotTerminated = do
             mTerminated <- readTVar terminated
             maybe (return ()) throwSTM mTerminated
+
+        getIsClosed :: STM (Maybe ClosedCode)
+        getIsClosed = do
+            st <- readTVar (streamState strm)
+            case st of
+                Closed code -> return $ Just code
+                _otherwise -> return Nothing
+
+        -- Check if the peer is still listening for messages
+        --
+        -- It is important to call 'checkNotClosed' prior to enqueuing stream
+        -- chunks to ensure that 'writeTBQueue' will not block indefinitely
+        -- (because nothing is consuming elements from the queue anymore).
+        --
+        -- Assumes 'checkNotTerminated'.
+        checkNotClosed :: STM ()
+        checkNotClosed = do
+            mClosed <- getIsClosed
+            case mClosed of
+                Just code ->
+                    -- When the stream is closed, but /we/ did not close it (or
+                    -- 'checkNotTerminated' would have thrown an exception), it
+                    -- must mean that our peer send us a RST_STREAM, indicating
+                    -- that they do not want to receive any further messages.
+                    throwSTM $ StreamRemoteReset code
+                _otherwise ->
+                    return ()
 
         iface :: OutBodyIface
         iface =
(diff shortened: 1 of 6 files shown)


Network/HTTP2/H2/Sync.hs
```diff
@@ -85,23 +84,29 @@ newLoopCheck strm mtbq = do
...
     }
 
 checkLoop :: LoopCheck -> IO Bool
 checkLoop LoopCheck{..} = atomically $ do
     tout <- readTVar lcTimeout
-    if tout
-        then return False
-        else do
+    state <- readTVar lcState
+    case (tout, state) of
```
kazu-yamamoto: I apologize if this is just a matter of personal preference, but please rewrite it using `MultiWayIf`.

Network/HTTP2/H2/OutBodyIface.hs
```diff
@@ -31,48 +33,86 @@ withOutBodyIface
...
+                _otherwise -> return Nothing
+
+        -- Check if the peer is still listening for messages
+        --
+        -- It is important to call 'checkNotClosed' prior to enqueuing stream
+        -- chunks to ensure that 'writeTBQueue' will not block indefinitely
+        -- (because nothing is consuming elements from the queue anymore).
+        --
+        -- Assumes 'checkNotTerminated'.
+        checkNotClosed :: STM ()
```
kazu-yamamoto, line 52: This code should move just after `checkNotTerminated` for readability.

Network/HTTP2/H2/OutBodyIface.hs
```diff
@@ -31,48 +33,86 @@ withOutBodyIface
...
+                _otherwise ->
+                    return ()
+
         iface :: OutBodyIface
         iface =
             OutBodyIface
                 { outBodyUnmask = unmask
                 , outBodyPush = \b -> atomically $ do
                     checkNotTerminated
+                    checkNotClosed
```
kazu-yamamoto, line 82: This guard itself represents a good idea.

This is a separate point from the change at hand, but have you considered what happens when an `RST_STREAM` is received while blocked in `writeTBQueue`?

Network/HTTP2/H2/Sync.hs
edsko: No need to apologize, a sense of aesthetics in programming is both very personal and very important :)

Network/HTTP2/H2/OutBodyIface.hs
edsko: Yes, I have, and it's exactly the code we're looking at here; this is the beauty of STM: a transaction such as

```haskell
checkNotTerminated
checkNotClosed
writeTBQueue tbq $ StreamingBuilder b NotEndOfStream
```

will be retried if _any_ of the variables it has read changes; so, if we are currently blocked in `writeTBQueue`, and we receive a `RST_STREAM`, the stream will be closed, the transaction will be retried, `checkNotClosed` will now fail, and the transaction will throw a `StreamRemoteReset <code>` exception.

edsko:
@kazu-yamamoto I've pushed a small commit that addresses your review feedback. Let me know if you prefer that I squash these changes into the older commits; I figured this way it would for you to see what I actually changed.

Network/HTTP2/H2/OutBodyIface.hs
kazu-yamamoto: I had always used `TBQueue` solely, so it had never even occurred to me that it could be composed using blocking `STM`. Excellent code!

kazu-yamamoto (approved):
Now LGTM

kazu-yamamoto:
Merged.

erikd:
@kazu-yamamoto Would be nice to get a Hackage release of this work from @edsko . This (I think) would allow Edsko release a new version of [grapesy](https://hackage.haskell.org/package/grapesy) which I am looking forward to because it fixes something in my dependency stack.

**Update:** Some exploration with Claude suggests that Edsko still has some new things that he needs in `http2`. Claude says:

>  This confirms conclusively (not just via source-grepping, but via an actual build attempt) that the API grapesy's 
> PR #375 needs isn't in http2 anywhere — not on Hackage, not on main, not even on the tip that already
> incorporates the maintainer's own most recent upstream contribution.

kazu-yamamoto:
@edsko The new release is up to you. Should I release the new version?

erikd:
@kazu-yamamoto I (or rather Claude) suspect that Edsko needs to add a couple of new things, but I will let him confirm.

edsko:
Morning/afternoon both :) I've been on PTO for a week. Yes, I think it would make sense to do a point release. I will be submitting a few more PRs against `http2` with additional fixes (HPACK bug due to received RST_STREAM, continuation frame handling for trailers, possibly a third), but none of those are regressions, so a point release would be a strict improvement (as far as I know) over the currently released version.

kazu-yamamoto:
v5.4.4 has been released.

edsko:
Thanks @kazu-yamamoto !

Issue #173, opened 2026-09-09
30 sec streaming timeout


goertzenator:
I have a web app that uses two long running SSE routes provided by a Servant server.  When using http/1.1 (`setHTTP2Disabled` for `Settings`) everything works fine, but when http/2 is enabled both of those fail after 30 seconds.

Below is my analysis.  Myself and Claude may have misread things so take this with a grain of salt.

The [timer on the worker thread is expiring](https://github.com/kazu-yamamoto/http2/blob/main/Network/HTTP2/Server/Worker.hs#L31).  The connection timer does get updated, but there is nothing that updates the worker timer and the stream dies after 30 sec.

In HTTP1 the timer is [paused and resumed while calling the application](https://github.com/yesodweb/wai/blob/master/warp/Network/Wai/Handler/Warp/HTTP1.hs#L189-L196).

This [does not happen in HTTP2.hs](https://github.com/yesodweb/wai/blob/master/warp/Network/Wai/Handler/Warp/HTTP2.hs#L98-L109).  The `aux0` variable is available and does contain the timer, so should pause/resume be added here?

kazu-yamamoto:
I think your analysis is correct. Please add `pause` and `resume` where you think they are needed, and if it looks good, submit a PR.

PR #174, opened 2026-09-11
worker timer is now tickled upon send

goertzenator:
This PR addresses issue #173 (30 second streaming timeout).

### Problem
An http2 stream in a Warp server that provides response data for more than the timeout (30 second default) will be killed by the worker timer.  This is common for a Server Sent Event route that might dribble events to a web client indefinitely.  I expect a plain file download that takes longer than the timeout period would also be killed.

### Solution
This PR adds a tickle to the worker timer whenever data is sent.  This keeps the worker alive while it is still sending data.

I have confirmed that this resolves the issue in the web app I am working on.


goertzenator:
Please don't merge this yet.  I am observing additional issues with my web app when sends are much less frequent.  I'd like to solve that problem too.

goertzenator:
I've decided to not fix the additional issues observed as I can't justify pouring too much time into this right now.  The fix in this PR is decent and is worth keeping IMO.

I did dig deeply into the way timers work so I will drop some notes in case myself or somebody else wants to take a look at things in the future.

To start I will lay out the timers involved in an http2 stream:
1. The Warp connection timer (external to http2).  The Warp HTTP2 module tickles this timer on any write and most reads (for slowloris protection).
2. There is an optional timer for the Receiver thread (one thread per connection)
3. There is a timer for the Worker (one per stream). This gets tickled on any send (this PR) or receive
4. There is a timer for the streaming body thread, if a streaming body is returned.  This is tickled when a chunk is provided.

While trying to understand timers in http2, I inspected how timers worked in Go and Rust implementations of http2.  Generally, those implementations default to not having timeouts.  The timeout options in the Go implementation seemed quite easy to understand, so I will walk through how those can be adapted to this Haskell http2 implementation.  

I struggled to understand the purpose of many of the timers in the Haskell implementation, and I think most of them are not needed.  The current effect is that they put timeout constraints on application code, something that Warp http/1.1 and the other language implementations do not do.  There might be concern about exposure to DOS attacks by removing some timers, but I will assert that the existing rate limit features to cover CVEs provide the real DOS protection.

| Setting | Default Value | Description |
| :--- | :--- | :--- |
| IdleTimeout | 0 = no timeout | Timeout for connection with no streams → closed with GOAWAY. "PING frames are not considered activity for the purposes of IdleTimeout." |
| ReadIdleTimeout | 0 = no health check | Nothing received → send a PING |
| PingTimeout | 0 → 15 s | PING unacked → close |
| WriteByteTimeout | 0 = no timeout | "Begins when data is available to write, and is extended whenever any bytes are written" |
| MaxConcurrentStreams | 0 → ≥100 | Per spec recommendation |


Design thoughts:
- The Warp connection timer should be placated with `pause` and then ignored.
- http2 should take all of the above settings as parameters.
- Remove all existing timers.
- Write timeouts can be implement in the Sender with `race (threadDelay writeByteTimeout) <operation>`.
- Alongside the Sender and Receiver threads, add another for idle monitoring.  It can monitor the stream tables to see when they are empty and run a timeout using the race approach above to watch for streams.
- Add yet another thread for read idle monitoring.  Sender can increment a TVar when a legit frame is received, and this thread will run a timeout on that and orchestrate a PING when needed.

kazu-yamamoto:
Just FYI: https://github.com/yesodweb/wai/pull/1109

kazu-yamamoto:
@goertzenator Please let me know your current thoughts.

Q1) Should this PR be merged as is?
Q2) Do you intend to implement the "Design thoughts" mentioned above?

goertzenator:
I think we should let the discussion in https://github.com/yesodweb/wai/pull/1109 play out before deciding anything here.

Issue #169, opened 2026-07-10
Build failure of `http-5.4.1` with `time-manager <0.3`.


phadej:
Signature of `withHandle` changed in `time-manager-0.3.0`.

```
% cabal build -c 'time-manager <0.3'
Build profile: -w ghc-9.2.8 -O1
In order, the following will be built (use -v for more details):
 - http2-5.4.1 (lib) (first run)
Preprocessing library for http2-5.4.1...
Building library for http2-5.4.1...
[40 of 48] Compiling Network.HTTP2.H2.Receiver ( Network/HTTP2/H2/Receiver.hs, dist-newstyle/build/x86_64-linux/ghc-9.2.8/http2-5.4.1/build/Network/HTTP2/H2/Receiver.o, dist-newstyle/build/x86_64-linux/ghc-9.2.8/http2-5.4.1/build/Network/HTTP2/H2/Receiver.dyn_o )

Network/HTTP2/H2/Receiver.hs:68:17: error:
    • Couldn't match type ‘Maybe Void’ with ‘Void’
      Expected: IO Void
        Actual: IO (Maybe Void)
    • In the expression:
        T.withHandle
          (threadManager ctx) (throwTo tid ConnectionIsTimeout) loop2
      In a stmt of a 'do' block:
        if confReadNTimeout then
            loop1
        else
            T.withHandle
              (threadManager ctx) (throwTo tid ConnectionIsTimeout) loop2
      In the expression:
        do labelMe "H2 receiver"
           tid <- myThreadId
           if confReadNTimeout then
               loop1
           else
               T.withHandle
                 (threadManager ctx) (throwTo tid ConnectionIsTimeout) loop2
   |
68 |                 T.withHandle (threadManager ctx) (E.throwTo tid ConnectionIsTimeout) loop2
   |  
```

http-5.4.0 compiles fine against older time-manager.

phadej:
I made a revision https://hackage.haskell.org/package/http2-5.4.1/revisions/, please apply it to the code as well (or make code work with older `time-manager` if preferred). Thanks!

kazu-yamamoto:
Done in v5.4.2.
Thanks!

PR #176, opened 2026-09-13
Restore simple-config timeout cleanup with time-manager <0.3

### LINKED ISSUE #175: freeSimpleConfig leaves timeout callbacks active with time-manager 0.3.x (cleanup issue)
We found this while investigating a `killManager` deprecation under a warning-fatal build policy on GHC 9.14.1/base 4.22. Checking what the deprecated call used to do exposed a behavioral difference at the public configuration lifecycle boundary.

`allocSimpleConfig'` creates a timeout manager exposed as `confTimeoutManager`. A callback registered with that manager can execute **after `freeSimpleConfig` returns** with time-manager 0.3.x. With HTTP2 5.4.0/time-manager 0.2.4, the same test observes no post-release callback.

### Related history

[Issue #169](https://github.com/kazu-yamamoto/http2/issues/169) reported that HTTP2 5.4.1 failed to compile with time-manager <0.3 because `withHandle` had changed its return type. The lower bound was corrected through a Hackage revision and [commit 96f9a77](https://github.com/kazu-yamamoto/http2/commit/96f9a77b99047fb49c65abfa408988ff0136ea2e), incorporated in HTTP2 5.4.2.

That bound correction addressed a real compilation incompatibility. However, it also made the manager series with no-op `killManager` mandatory, leaving the separate `freeSimpleConfig` callback-cleanup behavior described here unresolved. Simply lowering the bound again would therefore not be a sufficient fix for newer HTTP2 releases.

### Results

| HTTP2 | time-manager | No callback after free |
| --- | --- | --- |
| 5.4.0 | 0.2.4 | Pass during a 3-second observation |
| Each of 5.4.1, 5.4.2, 5.4.3, 5.4.4 | Each of 0.3.0, 0.3.1, 0.3.1.1, 0.3.2 | Fail on all 16 pairs: callback fires |

These are released package sources with an added test component, without HTTP2 or time-manager implementation patches. Runtime tests use a concrete resolved dependency plan for each pair; this is not an enumeration of all transitive dependency combinations.

### Related dependency/deprecation issue

The examined HTTP2 5.4.1–5.4.4 metadata requires `time-manager >=0.3.0 && <0.4`. All four releases in that range in our Hackage snapshot implement `killManager` as a deprecated no-op. `freeSimpleConfig` unconditionally calls it. Thus no otherwise buildable plan using those examined releases avoids that deprecated call when deprecation warnings are enabled. This claim is limited to the examined releases and metadata, not future versions.

The manager's semantic change is documented. The question is what HTTP2's allocation/free API should promise about registrations associated with its configuration. Simply deleting the deprecated call would remove the warning without restoring cancellation.

### Reproducer

Save the following as `test/regression/Main.hs` in an unpacked HTTP2 5.4.4 source tree:

```haskell
module Main (main) where
import Control.Concurrent (newEmptyMVar, putMVar, takeMVar)
import Control.Exception (bracket)
import Control.Monad (void)
import Network.HTTP2.Server (allocSimpleConfig', confTimeoutManager, freeSimpleConfig)
import qualified Network.Socket as S
import qualified System.TimeManager as T
import System.Timeout (timeout)
import Test.Hspec

main :: IO ()
main = hspec $ describe "simple configuration timeout ownership" $ do
  it "runs a callback before release (scheduler positive control)" $
    withPair $ \s -> bracket (allocSimpleConfig' s 4096 10000) freeSimpleConfig $ \c -> do
      fired <- newEmptyMVar
      void $ T.register (confTimeoutManager c) (putMVar fired ())
      timeout 2000000 (takeMVar fired) `shouldReturn` Just ()
  it "does not run a registered callback after freeSimpleConfig" $
    withPair $ \s -> do
      fired <- newEmptyMVar
      -- A one-second timer leaves time to free before expiry. The control
      -- above checks scheduler operation; absence is still bounded evidence.
      bracket (allocSimpleConfig' s 4096 1000000) freeSimpleConfig $ \c ->
        void $ T.register (confTimeoutManager c) (putMVar fired ())
      timeout 3000000 (takeMVar fired) `shouldReturn` Nothing

withPair :: (S.Socket -> IO a) -> IO a
withPair action = S.withSocketsDo $
  bracket (S.socket S.AF_INET S.Stream S.defaultProtocol) S.close $ \listener -> do
    S.bind listener (S.SockAddrInet 0 (S.tupleToHostAddress (127,0,0,1)))
    S.listen listener 1
    port <- S.socketPort listener
    bracket (S.socket S.AF_INET S.Stream S.defaultProtocol) S.close $ \client -> do
      S.connect client (S.SockAddrInet port (S.tupleToHostAddress (127,0,0,1)))
      bracket (fst <$> S.accept listener) S.close action

```

Add this component to `http2.cabal`:

```cabal
test-suite lifecycle-regression
  type: exitcode-stdio-1.0
  main-is: Main.hs
  hs-source-dirs: test/regression
  default-language: Haskell2010
  ghc-options: -Wall -threaded
  build-depends: base, http2, hspec, network, time-manager
```

Use a project containing that HTTP2 package and pin `time-manager ==0.3.2`, then run `cabal test http2:test:lifecycle-regression --test-show-details=direct`. For the passing historical control, use HTTP2 5.4.0 and time-manager 0.2.4 with the same test. Our expanded matrix also uses the released manager source trees as local packages to select each exact version.

The failing assertion is:

```text
expected: Nothing
 but got: Just ()
2 examples, 1 failure
```

The control checks that a live manager delivers a callback. The absence assertion is bounded to three seconds; it is not proof of indefinite absence. A delivered callback after release is direct evidence of the changed behavior. This test does not demonstrate a production use-after-free, data corruption, or a leak in an actual HTTP/2 request path.

### Current workaround and proposed discussion

We pin **both** HTTP2 5.4.0 and time-manager 0.2.4 and retain the lifecycle regression. Pinning HTTP2 5.4.0 alone is insufficient because its manager range also admits 0.3.x.

Should `freeSimpleConfig` retain the earlier cancellation behavior, or should registrations have a different explicit ownership/release contract? We can contribute the regression test as a test-only PR once that expected contract is agreed.

gorban:
Restore the pending-timeout cleanup performed by `freeSimpleConfig` by requiring `time-manager >=0.2.4 && <0.3`, and adapt the receiver to manage its timeout handle with `bracket register cancel`. Fixes #175.

### Why the bound change also needs code

#169 correctly identified that older `withHandle` returns `IO (Maybe a)`, whereas the receiver needs `IO Void`; raising the lower bound fixed that compilation error. However, time-manager 0.3 also made `killManager` a deprecated no-op, so pending registrations can remain active after configuration release.

Explicit handle bracketing avoids the return-type mismatch without swallowing timeout exceptions. It allows the older manager's cleanup behavior to be retained. The lower bound is 0.2.4 because current HTTP2 uses `isAllGone`, introduced in that release.

**Compatibility tradeoff:** this excludes time-manager 0.3 and retains the older manager implementation. Applications requiring 0.3 in the same dependency plan cannot use this proposal unchanged. A solution retaining 0.3 would need an agreed replacement registration-ownership contract or manager support; this PR proposes the smaller restoration of the earlier behavior.

### Verification

Linux, GHC 9.10.3, current main `d971792`:

| Check | Result |
| --- | --- |
| Current implementation with time-manager 0.3.2 | Live-callback and explicit-handle cancellation controls pass; post-free regression fails |
| Change only the bound to use 0.2.4 | Reproduces #169: `Maybe Void` does not match `Void` |
| Bound plus explicit receiver-handle bracketing, with `-Werror` | All 41 examples pass across all three enabled suites |
| Receiver after a valid preface followed by no frames | Propagates `ConnectionIsTimeout`, as expected |

The first commit adds three lifecycle cases to the existing `spec` suite; the fix retains them and adds the receiver-timeout case. Before the fix, `spec` has 38 examples with exactly one regression failure; `hpack` and `frame` pass. Afterward, `spec` has 39 examples and both other suites pass. No new test suite, fixtures, warning suppression or removed assertions. Existing CI runs these suites via `cabal test ... all`; its optional `h2spec` component stays disabled by the existing default flag.

The post-free assertion observes three seconds, so absence is bounded evidence. It is not a demonstrated production request-path use-after-free. Existing issue #175 also records the released-version comparison. Other compilers/OSes remain for CI after submission.

Commands used `cabal test all --keep-going --test-show-details=direct`, with an external project pinning the compiler, index state and exact time-manager version; the passing project sets `package http2: ghc-options: -Werror`. Full commands, logs and plans are available. The changelog records this proposal under Unreleased and links this PR; a release version can be assigned once the compatibility direction is accepted.


kazu-yamamoto:
In my opinion, this should be fixed in the upstream `timer-manager`. I have successfully imported the `ConfigSpec` you created into `timer-manager`, so I will now consider whether a fix is ​​possible.

kazu-yamamoto:
My current view is as follows:

- Downgrading the dependency to `time-manager < 0.3` is unacceptable.
- `register` should be used in conjunction with `cancel` within a `bracket`. Since `register` is a low-level function, `withHandle` should generally be used instead. It might be advisable to emphasize in the `register` documentation that cancellation is the user's responsibility.

